PR #2, opened 2021-01-13
DM-28148: add color color plots

sr525:


python/lsst/analysis/drp/qaUtils.py
```diff
@@ -0,0 +1,225 @@
...
+
+    for band in ["g", "r", "i", "z", "y"]:
+        usePerBand = ((cat[band + "Centroid_flag_notAtMaximum"] == 0) & (cat[band + "Shape_flag"] == 0)
+                      & (cat[band + "PsfFlux_flag"] == 0)
+                      & (cat[band + "PixelFlags_saturatedCenter"] == 0)
+                      & (cat[band + "Extendedness_flag"] == 0)
+                      & (cat[band + "Deblend_nChild"] == 0))
+        use = use & usePerBand
+    cat["useForQAFlag"] = use
+    return cat
```
timothydmorton, line 153: What about something like this?
```
use = np.ones(len(cat)).astype(bool)

for b in bands:  # add bands as a keyword arg above
    for magType in magTypes:  # add magTypes as a kwarg too
        use &= np.isfinite(cat[b + magType + 'Mag'])
    for flag in flagList:  # make flagList a kwarg
        use &= cat[band + flag] == 0

    cat.loc[:, 'useForQAFlag'] = use
    return cat
```

timothydmorton: keeping with the general pattern to make all these more customizable if needed ever...  which makes me wonder...why aren't all these functions actions?  Seems like a general action could`appendAllTheColumns`.  Probably not necessary, but just wondering why the actions stop here.

python/lsst/analysis/drp/qaUtils.py
```diff
@@ -0,0 +1,225 @@
...
+    cat : `pandas.core.frame.DataFrame`
+    fluxColName : `str`
+        Optional, defaults to "PsFlux"
+
+    Returns
+    -------
+    cat : `pandas.core.frame.DataFrame`
+    """
+
+    for band in ["g", "r", "i", "z", "y"]:
```
timothydmorton, line 170: `bands` kwarg

python/lsst/analysis/drp/qaUtils.py
```diff
@@ -0,0 +1,225 @@
...
+        SN = cat[band + fluxColName].values / cat[band + fluxColName + "Err"].values
+        cat[band + "Sn" + fluxColName] = SN
```
timothydmorton, line 172: same as above.  no `.values` and use `.loc[:, colName]`.

python/lsst/analysis/drp/qaUtils.py
```diff
@@ -0,0 +1,225 @@
...
+    Returns
+    -------
+    cat : `pandas.core.frame.DataFrame`
+     """
+
+    # TODO: Is this needed? Where do the thresholds come from?
+    # Can potentially get rid of this as S/N is needed
+    # as its own column for other plots
+    useForStats = np.zeros(len(cat))
+    SN = cat["iPsFlux"].values / cat["iPsFluxErr"].values
```
timothydmorton, line 194: `.values`

python/lsst/analysis/drp/qaUtils.py
```diff
@@ -0,0 +1,225 @@
...
+
+    lowSn = np.where((SN > 500))[0]
+    useForStats[lowSn] = 2
+
+    highSn = np.where((SN > 2700))[0]
+    useForStats[highSn] = 1
```
timothydmorton, line 200: `i` band, `500` and `2700` are all hardcoded in this function.  From the comments above it looks like maybe this really isn't being used, but just pointing out.  And again, I thought there was some action that was taking care of the useForStats selection?

python/lsst/analysis/drp/qaUtils.py
```diff
@@ -0,0 +1,225 @@
...
+    -------
+    cat : `pandas.core.frame.DataFrame`
+    """
+    # TODO: Is this now these should be calculated?
+
+    for band in ["g", "r", "i", "z", "y"]:
+        shape = cat[band + "ShapeRound_xx"].values + cat[band + "ShapeRound_yy"].values
+        psfShape = cat[band + "IxxPsf"].values + cat[band + "IyyPsf"].values
+        cat[band + "DeconvMoments"] = shape
+        cat[band + "PsDeconvMoments"] = psfShape
```
timothydmorton, line 223: `.values` and `.loc[:, ...]`

python/lsst/analysis/drp/addQAColumns.py
```diff
@@ -0,0 +1,65 @@
...
+                                            name="qaTable_tract",
+                                            dimensions=("tract", "skymap"))
+
+
+class AddQAColumnsTaskConfig(pipeBase.PipelineTaskConfig, pipelineConnections=AddQAColumnsTaskConnections):
+
+    pass
+
+
+class AddQAColumnsTask(pipeBase.PipelineTask):
```
timothydmorton, line 27: I'd like this to be named `AppendQAColumnsTask` (as well as `add` --> `append` in many other places.  "add" suggests to me a mathematical operation, whereas appending columns is what this is doing.

timothydmorton (changes requested):
Mostly small pandas-related comments and minor suggestions, but one bigger-picture comment is that I don't understand why a lot of the functions in `qaUtils.py` are functions there instead of being implemented more generally as `DataFrameAction`s that could also be used elsewhere (e.g., in a `TransformTableTask` or something), as well as in `AppendQAColumnsTask`.  I understand it might be overkill, but things like appending mag columns, calculating deconvolvedMoments, etc. seem like they should be more general-purpose than util functions in an afterburner analysis package.  e.g., if we wanted to run this for bands other than `grizy`, then it would be useful to get that info from somewhere upstream and make sure all these appending column things use that.  And it might be useful for the flag choices in the `useForQA` stuff to be configurable, too.  Though that could also happen just at the `AppendQAColumnsTaskConfig` level, and then passing the necessary kwargs on to the various appending functions, without needing to turn everything into actions.  But it just seems consistent that if most of the operations on dataframes we're doing are actions (like the selectors, etc), then those could be, too.

Then there's one of the yaml files that currently has a merge conflict in it; I'll approve after that's fixed and I can take a look.

pipelines/colorColorPlotTest.yaml
```diff
@@ -0,0 +1,28 @@
+description: Make colour analysis plots
+tasks:
+  addQAColumns:
+    class: lsst.analysis.drp.addQAColumns.AddQAColumnsTask
+  plot_ri_gr_CModel:
+    class: lsst.analysis.drp.colorColorPlot.ColorColorPlotTask
+    config:
+      xColName: grCModelMag
```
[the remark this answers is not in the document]
sr525: This file should not have been included, nor should skyPlotExample. I will dig into what happened.

timothydmorton (approved):
My comments about preferring "append" to "add" for the tasks/functions(/actions?) that append columns to dataframes still hold, even if `addQAColumns.py` wasn't supposed to be in this PR.

python/lsst/analysis/drp/colorColorPlot.py
```diff
@@ -0,0 +1,208 @@
...
+                   self.config.axisLabels["z"]: self.config.axisActions.zAction(catPlot),
+                   "patchId": catPlot["patchId"]}
+        plotDf = pd.DataFrame(columns)
+
+        sourceTypes = np.zeros(len(plotDf))
+        for selector in self.config.sourceSelectorActions:
+            # The source selectors return 1 for a star and 2 for a galaxy
+            # rather than a mask this allows the information about which
+            # type of sources are being plotted to be propagated
+            sourceTypes += selector(catPlot)
```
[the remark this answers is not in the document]
sr525: I wanted to do it this way so that if we wanted to add other types later we could without having to add tons of columns.

python/lsst/analysis/drp/colorColorPlot.py
```diff
@@ -0,0 +1,208 @@
...
+        if list(self.config.sourceSelectorActions) == []:
+            sourceTypes = [10]*len(plotDf)
+        plotDf.loc[:, "sourceType"] = sourceTypes
+
+        # Check the columns have finite values
+        mask = np.ones(len(catPlot), dtype=bool)
+        for col in plotDf.columns:
+            mask &= np.isfinite(plotDf[col])
+        plotDf = plotDf[mask]
```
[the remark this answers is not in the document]
sr525: Thank you for the advice, I find the current version easier to understand. One day maybe I will learn to speak pandas.

python/lsst/analysis/drp/qaUtils.py
```diff
@@ -0,0 +1,225 @@
+import numpy as np
+
+
+def addMagnitudes(cat, fluxColName="PsFlux"):
+    """Add magnitude columns.
```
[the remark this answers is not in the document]
sr525: This file is also outdated and needs to be deleted.

timothydmorton (approved):

PR #4, opened 2021-01-14
DM-28149: Add a task to make scatter plots

sr525:


python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -0,0 +1,532 @@
...
+                                                  storageClass="Plot",
+                                                  name="scatterTwoHistPlot_{plotName}",
+                                                  dimensions=("tract", "skymap"))
+
+
+class ScatterPlotWithTwoHistsTaskConfig(pipeBase.PipelineTaskConfig,
+                                        pipelineConnections=ScatterPlotWithTwoHistsTaskConnections):
+
+    axisActions = ConfigurableActionStructField(
+        doc="The actions to use to calculate the values used on each axis.",
```
timothydmorton, line 47: Maybe explain here what the "mag" axis is?  `xAction` and `yAction` seem self-explanatory, but I don't understand off the bat what `magAction` is.

python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -0,0 +1,532 @@
...
+        default={"mainFlagSelector": dataSelectors.MainFlagSelector},
+    )
+
+    highSnStatisticSelectorActions = ConfigurableActionStructField(
+        doc="Selectors to use to decide which points to use for calculating statistics.",
+        default={"statSelector": dataSelectors.HighSnSelector},
+    )
+
+    lowSnStatisticSelectorActions = ConfigurableActionStructField(
+        doc="Selectors to use to decide which points to use for calculating statistics.",
```
timothydmorton, line 69: doc strings here and above should distinguish what the difference is between `highSnStatisticSelectorActions` and `lowSnStatisticSelectorActions`; currently their descriptions are identical.

python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -0,0 +1,532 @@
...
+    def scatterPlotWithTwoHists(self, catPlot, plotInfo, sumStats, yLims=False, xLims=False):
+
+        """Makes a generic plot with a 2D histogram and collapsed histograms of
+        each axis.
+
+        Parameters
+        ----------
+        catPlot : `pandas.core.frame.DataFrame`
+            The catalog to plot the points from.
+        xLabel : `str`
```
timothydmorton, line 211: update docstring to reflect actual parameters. `xLabel` and `yLabel` are not here.

python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -0,0 +1,532 @@
...
+            If being given a tuple of (xMin, xMax).
+
+        Returns
+        -------
+        fig : `matplotlib.figure.Figure`
+            The resulting figure.
+
+        Notes
+        -----
+        Uses the config options `xColName` and `yColName` to plot a scatter
```
timothydmorton, line 245: I think this info is a bit old, too, predating the "actions" updates?

python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -0,0 +1,532 @@
...
+            highStats += ["Median: {:0.2f}".format(highSnGalMed),
+                          r"$\sigma_{MAD}$: " + "{:0.2f}".format(highSnGalMad)]
+            lowStats += ["Median: {:0.2f}".format(lowSnGalMed),
+                         r"$\sigma_{MAD}$: " + "{:0.2f}".format(lowSnGalMad)]
+
+        if np.any(catPlot["sourceType"] == 1):
+            highSnStars = ((catPlot["useForStats"] == 1) & stars)
+            highSnStarMed = np.nanmedian(catPlot.loc[highSnStars, yCol])
+            highSnStarMad = sigmaMad(catPlot.loc[highSnStars, yCol], nan_policy="omit")
+            approxHighMag = np.nanmax(catPlot.loc[highSnStars, magCol])
```
timothydmorton, line 295: OK, I think I see the role of `magCol` now. It's to allow any generic plot with any x column, and any generic selection for stats, but then, whatever that is, to always report a mag here instead of a value of `x`?  It seems a bit specialized.  Might help with understanding its role to also have an example demo plot that shows a case where `magCol` and `xCol` are not the same, but where `magCol` would be missed if it didn't exist.

python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -0,0 +1,532 @@
...
+                 horizontalalignment="center", verticalalignment="center", fontsize=8)
+
+        plt.draw()
+
+        # Add a legend
+        for (i, newLabel) in enumerate(highStats):
+            linesForLegend[i].set_label(newLabel)
+        highStatThresh = self.config.highSnStatisticSelectorActions.statSelector.threshold
+        lowStatThresh = self.config.lowSnStatisticSelectorActions.statSelector.threshold
+        highLegendTitle = "S/N > {} Stats \n({} < {:0.2f})".format(highStatThresh, xCol, approxHighMag)
```
timothydmorton, line 515: Wait- but this line now conflicts with my understanding of the difference between `xCol` and `magCol`.  `approxHighMag` was computed from the `magCol` column, but now the legend is reporting as `xCol` less than something.  I think either `xCol` here should be `magCol`, or `magCol` itself should just become `xCol` (as I said above, perhaps an example illustrating an example of the need for `magCol` in addition to `xCol` would be instructive).

python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -0,0 +1,532 @@
...
+        fig.legend(handles=linesForLegend, fontsize=6, bbox_to_anchor=(0.81, 0.655), edgecolor="k",
+                   bbox_transform=fig.transFigure, title=highLegendTitle, title_fontsize=6)
+        for (i, newLabel) in enumerate(lowStats):
+            linesForLegend[i].set_label(newLabel)
+        lowLegendTitle = "S/N > {} Stats \n({} < {:0.2f})".format(lowStatThresh, xCol, approxLowMag)
```
timothydmorton, line 520: ditto above (xCol in title whereas magCol used in calculation).

pipelines/skyPlotExample.yaml
```diff
@@ -81,3 +81,34 @@ tasks:
...
         from lsst.analysis.drp.dataSelectors import PsfFlagSelector, HighSnSelector, GalaxyIdentifier
         from lsst.analysis.drp.calcFunctors import SNCalculator
+
+ # Same as above but this time plot it for galaxies not stars
+ # The defualt is stars but you can also use all for both.
+ # Then adjust the SN threshold for points to be included in the plot.
+  plot_psfSn_goodCmodel_stars_scatter:
+    class: lsst.analysis.drp.scatterPlot.ScatterPlotWithTwoHistsTask
+    config:
+      axisActions.yAction: SNCalculator
```
timothydmorton, line 91: I feel like somewhere in this example (perhaps in the comment preceding this block) it should explain that there is a default `xAction` and what it is, and that's why no `xAction` is provided on this scatterplot example.

python/lsst/analysis/drp/scatterPlot.py
timothydmorton: Also seems like the fact that there is actually a default xAction should be mentioned in the doc here, since right now it's just hidden in the `setDefaults` method.

timothydmorton (approved):
All looks generally OK here, just a few minor suggestions, mostly about documentation.  One additional soft suggestion that does not relate to any code specifically but that I mention a few times and will flesh out here: I think it might be instructive to run an example of this kind of plot for something other than a mag as the x axis.  To me, this would help make clear whether a distinction really is necessary between `xCol` and `magCol`, and also whether there are any other built-in gotchas that come from all of the usage to date being with mag as x-col.  In other words, this is built as a general x-y scatter plot thing, but it might work actually less generally than intended (especially with the drawing of the paths, etc.), in which case maybe we should be content with this being just more narrowly a MagScatterHistPlot or something.

python/lsst/analysis/drp/scatterPlot.py
sr525: The mag axis is used to decide on the points which go into the statistics.

python/lsst/analysis/drp/scatterPlot.py
sr525: I've added an example that covers both this and the points you made on the jira ticket.

python/lsst/analysis/drp/scatterPlot.py
sr525: Good catch, that should be magCol.

PR #5, opened 2021-01-14
DM-28144: Add tasks to calculate and plot the stellar locus fits.

sr525:


python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -0,0 +1,385 @@
...
+        lineRms = axHist.axvline(meanDists + rmsDists, color="C0", ls="--",
+                                 label="RMS: {:0.3f}".format(rmsDists))
+        axHist.axvline(meanDists - rmsDists, color="C0", ls="--")
+
+        linesForLegend = [lineHW, lineInit, lineRefit, fitScatter, lineMedian, lineMad, lineMean, lineRms]
+        fig.legend(handles=linesForLegend, fontsize=8, bbox_to_anchor=(1.0, 0.99),
+                   bbox_transform=fig.transFigure, ncol=2)
+
+        axHist.hist(dists, bins=100, histtype="step", label="Refit", color="C0")
+        axHist.hist(distsHW, bins=100, histtype="step", label="HW", color="C0", alpha=0.5)
```
jeffcarlin, line 359: Rather than a fixed number of bins, I generally prefer a fixed bin _width_ so that histograms are easier to interpret. For example, you could say "bins=np.arange(xmin, xmax, binsize)", with binsize=0.01 or something. (If it's too hard, it's probably not vitally important in this case, though...)

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -105,3 +107,163 @@ def addPlotInfo(fig, plotInfo):
...
+            The intercept from the second round of fitting
+        mPerp : `float`
+            The gradient of the line perpendicular to the line from the
+            second fit
+
+    Notes
+    -----
+    The code does two rounds of fitting, the first is initiated using the
+    hardwired values given in the `paramDict` parameter and is done using
+    an ODR fit to the points defined by the box of xMin, xMax, yMin and
```
jeffcarlin, line 176: What is ODR? (Spell out the 3 words...)

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -105,3 +107,163 @@ def addPlotInfo(fig, plotInfo):
...
+
+    # Points to use for the fit
+    fitPoints = np.where((xs > paramDict["xMin"]) & (xs < paramDict["xMax"])
+                         & (ys > paramDict["yMin"]) & (ys < paramDict["yMax"]))[0]
+
+    def linearFit(B, x):
+        return B[0]*x + B[1]
+
+    linear = scipyODR.Model(linearFit)
+    linear = scipyODR.polynomial(1)
```
jeffcarlin, line 190: Don't you want to choose one or the other? This appears to be defining "linear" twice.

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -105,3 +107,163 @@ def addPlotInfo(fig, plotInfo):
...
+
+    paramsOut["yBoxMin"] = yBoxMin
+    paramsOut["bPerpMin"] = bPerpMin
+
+    if mODR > 5:
+        yBoxMax = paramDict["yMax"]
+        xBoxMax = (yBoxMax - bODR)/mODR
+    else:
+        yBoxMax = mODR*paramDict["xMax"] + bODR
+        xBoxMax = paramDict["xMax"]
```
jeffcarlin, line 223: For simplicity/readability, you could combine these into a single if/else statement, instead of two separate ones.

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -105,3 +107,163 @@ def addPlotInfo(fig, plotInfo):
...
+    paramsOut["yBoxMax"] = yBoxMax
+    paramsOut["bPerpMax"] = bPerpMax
+
+    # Use these perpendicular lines to chose the data and refit
+    fitPoints = ((ys > mPerp*xs + bPerpMin) & (ys < mPerp*xs + bPerpMax))
+    data = scipyODR.Data(xs[fitPoints], ys[fitPoints])
+    odr = scipyODR.ODR(data, linear, beta0=[bODR, mODR])
+    params = odr.run()
+    mODR = float(params.beta[1])
+    bODR = float(params.beta[0])
```
jeffcarlin, line 236: Do you need some checks here (and when fitting above) to throw an error (or at least warn) when the fitting fails?

jeffcarlin (changes requested):
This is really nice, Sophie. I have a few comments/questions, though:

1. The biggest thing is that I think you've got the wrong column names for all of the fluxes. 
2. The stellar locus width should really always be calculated with extinction-corrected magnitudes. It looks like the actual stellar locus fitting/calculation is agnostic about its inputs (i.e., it just expects "x" and "y"), so that enables whatever corrections you choose before calling that function. But I would recommend adding extinction corrections to the routine fitting/plotting pipelines.
3. Similar to extinction, we would also like to apply external photocalibs (e.g., FGCM) before calculating the stellar locus width. Is that possible in the current setup? [Though I think that eventually the FGCM corrections will be applied to fluxes when creating ObjectTables, so maybe we just have to wait?]
4. Have you tried using "plt.minorticks_on()"? I always find plots are easier to read with the intermediate tick marks.
5.

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -0,0 +1,385 @@
...
+        mask = np.ones(len(catPlot), dtype=bool)
+        for actionStruct in [self.config.selectorActions, self.config.sourceSelectorActions]:
+            for selector in actionStruct:
+                mask &= selector(catPlot)
+        catPlot = catPlot[mask]
+
+        columns = {self.config.axisLabels["x"]: self.config.axisActions.xAction(catPlot),
+                   self.config.axisLabels["y"]: self.config.axisActions.yAction(catPlot),
+                   self.config.axisLabels["mag"]: self.config.axisActions.magAction(catPlot),
+                   "patchId": catPlot["patchId"]}
```
jeffcarlin, line 111: I don't understand why patchId is being passed here. Why is this needed in the `runQuantum` method?

sr525:
> This is really nice, Sophie. I have a few comments/questions, though:
> 
> 1. The biggest thing is that I think you've got the wrong column names for all of the fluxes.
> 2. The stellar locus width should really always be calculated with extinction-corrected magnitudes. It looks like the actual stellar locus fitting/calculation is agnostic about its inputs (i.e., it just expects "x" and "y"), so that enables whatever corrections you choose before calling that function. But I would recommend adding extinction corrections to the routine fitting/plotting pipelines.
> 3. Similar to extinction, we would also like to apply external photocalibs (e.g., FGCM) before calculating the stellar locus width. Is that possible in the current setup? [Though I think that eventually the FGCM corrections will be applied to fluxes when creating ObjectTables, so maybe we just have to wait?]
> 4. Have you tried using "plt.minorticks_on()"? I always find plots are easier to read with the intermediate tick marks.
> 5. Is "plotUtils.py" the right place for the "stellarLocusFit" function? It's not really a "plot" util.

1) Opps, I'll check again.
2 and 3) I want to make these plots from what is available to end users, currently extinction correction isn't applied to these. I will make another ticket to look into this (DM-32791). What FGCM corrections are you hoping to apply? Aren't the fluxes already calibrated?
4) I prefer plots without minor ticks but I can turn them on if people find this useful.
5) If other people are going to use it then it can be moved to a different file. Currently it is only used by the plotting code.

jeffcarlin:
> 1. 2 and 3) I want to make these plots from what is available to end users, currently extinction correction isn't applied to these. I will make another ticket to look into this (DM-32791). What FGCM corrections are you hoping to apply? Aren't the fluxes already calibrated?

For (2): I think extinction correction is essential for the stellar locus to be meaningful. (I get your point re: end products available to users, but the stellar locus really needs to be corrected to be useful.) There's an extinction correction module in faro ([here](https://github.com/lsst/faro/blob/main/python/lsst/faro/utils/extinction_corr.py)), but it assumes afw tables as inputs. A parquet-compatible version is coming soon.

For (3), it is my understanding that the FGCM calibrations have _not_ been applied to values in the objectTables. (Yusra tells me that this _will_ be true some day soon, but is not true yet...)

jeffcarlin:
> * I prefer plots without minor ticks but I can turn them on if people find this useful.
> * If other people are going to use it then it can be moved to a different file. Currently it is only used by the plotting code.

No big deal about the minor ticks - I just find it makes reading values from plots easier.

I imagine the stellar locus fitting could be useful elsewhere (faro?), so it might make sense for it to live somewhere that isn't labeled "plots".

pipelines/stellarLocusPlots.yaml
```diff
@@ -0,0 +1,155 @@
...
+                   "y": "r - i PS Magnitude (mag)",
+                   'mag': "r PS Magnitude (mag)"}
+      selectorActions.flagSelector: CoaddPlotFlagSelector
+      selectorActions.flagSelector.bands: ["g", "r", "i"]
+      selectorActions.SnSelector: SnSelector
+      selectorActions.SnSelector.fluxType: "PsFlux"
+      selectorActions.SnSelector.threshold: 50
+      selectorActions.SnSelector.bands: ["g", "r", "i"]
+      stellarLocusFitDict: {"xMin": 0.28, "xMax": 1.0, "yMin": 0.02, "yMax": 0.48,
+                            "mHW": 0.52, "bHW": -0.08}
```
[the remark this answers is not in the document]
sr525: They are cuts in the colors relevant for the specific fits taken from pipe_analysis. I've added a comment to the pipeline file.

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -0,0 +1,385 @@
...
+    stellarLocusFitDict = pexConfig.DictField(
+        doc="The parameters to use for the stellar locus fit.",
+        keytype=str,
+        itemtype=float,
+    )
+
+    fitName = pexConfig.Field(
+        doc="The name of the fit.",
+        dtype=str,
+    )
```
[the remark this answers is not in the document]
sr525: This can just be removed it is outdated.

python/lsst/analysis/drp/plotUtils.py
sr525: The plot will make it obvious if the fit has failed but produced something, otherwise I think it will crash.

sr525:
> > 1. 2 and 3) I want to make these plots from what is available to end users, currently extinction correction isn't applied to these. I will make another ticket to look into this (DM-32791). What FGCM corrections are you hoping to apply? Aren't the fluxes already calibrated?
> 
> For (2): I think extinction correction is essential for the stellar locus to be meaningful. (I get your point re: end products available to users, but the stellar locus really needs to be corrected to be useful.) There's an extinction correction module in faro ([here](https://github.com/lsst/faro/blob/main/python/lsst/faro/utils/extinction_corr.py)), but it assumes afw tables as inputs. A parquet-compatible version is coming soon.
> 
> For (3), it is my understanding that the FGCM calibrations have _not_ been applied to values in the objectTables. (Yusra tells me that this _will_ be true some day soon, but is not true yet...)

We aren't going to need to apply some extra FGCM stuff to the object table outputs. Currently it is used to calibrate things in the processing, if we end up using more FGCM calibrations they will be applied before the object table point.

I have made a ticket to add the extinction correction to the object tables. Doing that on this ticket would be too much scope creep.

sr525:
> > * I prefer plots without minor ticks but I can turn them on if people find this useful.
> > * If other people are going to use it then it can be moved to a different file. Currently it is only used by the plotting code.
> 
> No big deal about the minor ticks - I just find it makes reading values from plots easier.
> 
> I imagine the stellar locus fitting could be useful elsewhere (faro?), so it might make sense for it to live somewhere that isn't labeled "plots".

I think that there is/will be a bunch of code that should be shared between the two and that we should figure out where to put it and move everything to that place.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -10,17 +10,17 @@ class SNCalculator(DivideColumns):
 
     def setDefaults(self):
         super().setDefaults()
-        self.colA.column = "iPsFlux"
-        self.colB.column = "iPsFluxErr"
+        self.colA.column = "i_PsfFlux"
+        self.colB.column = "i_PsfFluxErr"
```
jeffcarlin, line 14: Should be `i_psfFlux` and `i_psfFluxErr` (lower case "p").

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -10,17 +10,17 @@ class SNCalculator(DivideColumns):
...
 
 class KronFluxDivPsfFlux(DivideColumns):
     """Divide the Kron instFlux by the PSF instFlux"""
 
     def setDefaults(self):
         super().setDefaults()
-        self.colA.column = "iKronFlux"
-        self.colB.column = "iPsFlux"
+        self.colA.column = "i_KronFlux"
+        self.colB.column = "i_PsfFlux"
```
jeffcarlin, line 23: Should be `i_kronFlux` and `i_psfFlux` (lower case "k" and "p").

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -0,0 +1,378 @@
...
+        default={"xAction": MagDiff, "yAction": MagDiff, "magAction": MagColumnNanoJansky},
+    )
+
+    axisLabels = pexConfig.DictField(
+        doc="Axis labels for the plot.",
+        keytype=str,
+        itemtype=str,
+        default={"x": "g - r PS Magnitude (mag)",
+                 "y": "r - i PS Magnitude (mag)",
+                 "mag": "r PS Magnitude (mag)"},
```
jeffcarlin, line 51: Should these say "PSF magnitude" instead of "PS"?

python/lsst/analysis/drp/histPlot.py
```diff
@@ -0,0 +1,406 @@
...
+        "is plotted per-panel, the percentile limit is the maximum value across all input data.",
+        dtype=float,
+        default=98.0,
+    )
+
+    nBins = pexConfig.Field(
+        doc="Number of bins used to divide the x axes.",
+        default=40,
+        dtype=int,
+    )
```
jeffcarlin, line 111: As I commented on colorColorFitPlot.py, it is useful to be able to explicitly set the bin width rather than the number of bins. It's easier to read and compare histograms when you know the width of the bins.

python/lsst/analysis/drp/histPlot.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def setDefaults(self):
+        super().setDefaults()
+        self.leftPanelActions.hist1.column = "iPsFlux"
+        self.leftPanelActions.hist2.column = "iAp9Flux"
+        self.rightPanelActions.hist1.colA.column = "iPsFlux"
+        self.rightPanelActions.hist1.colB.column = "iPsFluxErr"
+        self.rightPanelActions.hist2.colA.column = "iAp9Flux"
+        self.rightPanelActions.hist2.colB.column = "iAp9FluxErr"
```
jeffcarlin, line 120: Incorrect column names.

jeffcarlin (approved):
There are a couple of comments from my first review that didn't get addressed, and some renamings of columns got missed (including throughout `coaddQAPlots.yaml`). I added some (small) new columns, but otherwise I think it's looking pretty good.

In the absence of unit tests, I would suggest making at least one plot that exercises each functionality of each of the plot types, to make sure that there are no bugs or wrong column names still hanging around. Then I think it's good to merge.

sr525:
> There are a couple of comments from my first review that didn't get addressed, and some renamings of columns got missed (including throughout `coaddQAPlots.yaml`). I added some (small) new columns, but otherwise I think it's looking pretty good.
> 
> In the absence of unit tests, I would suggest making at least one plot that exercises each functionality of each of the plot types, to make sure that there are no bugs or wrong column names still hanging around. Then I think it's good to merge.

Sorry I accidentally included a commit that is already merged, think I forgot to rebase to main. The column names that are wrong in this and the calcFunctors bits that aren't executed by this pipeline will be fixed on a different ticket. Tests are also being added before too long on another ticket.

PR #6, opened 2021-01-22
DM-28389: Create a gen3 task to collate tract-level parquet tables for QA analyses

laurenam:


python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        if self.config.filterMap is not None:
+            if band in self.config.filterMap:
+                filterName = self.config.filterMap[band]
+            else:
+                raise RuntimeError("A config.filterMap was provided, but there is no entry for the "
+                                   "band requested: {}".format(band))
+        else:
+            filterName = band
+        self.log.info("Aggretating QA table for {} tract {} including {} patches".
```
TallJimbo, line 204:
```suggestion
        self.log.info("Aggregating QA table for {} tract {} including {} patches".
```

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,284 @@
...
+                      format(filterName, tract, len(inputObjCats)))
+        # It is much faster to concatenate a list of DataFrames than to
+        # concatenate successively within the for loop.
+        catList = []
+        colsToLoadList = None
+        refColsToLoadList = None
+        measColsToLoadList = None
+        for inputCat in inputObjCats:
+            patch = inputCat.dataId["patch"]
+            inputCat = inputCat.get()  # Requied because deferLoad=True was set
```
TallJimbo, line 214:
```suggestion
            inputCat = inputCat.get()  # Required because deferLoad=True was set
```

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,284 @@
...
+            ``band``.
+        """
+        if self.config.filterMap is not None:
+            if band in self.config.filterMap:
+                filterName = self.config.filterMap[band]
+            else:
+                raise RuntimeError("A config.filterMap was provided, but there is no entry for the "
+                                   "band requested: {}".format(band))
+        else:
+            filterName = band
```
TallJimbo, line 203: Now that I see how you're using `filterMap`, I don't actually understand why you'd want to: you're starting with a `band` value that's what is directly used to identify the coadd (note that in Gen3, a coadd data ID uses `band`, _not_ `physical_filter`, because multiple `physical_filter` values can and often do go into one coadd).

So, by switching to `physical_filter`-like names, you're actually making the plots lie sometimes: they're saying that a coadd that includes one or more HSC-R2 visits is an HSC-R coadd, and that's not true.  I'm sure this dates back to before we started trying to be consistent about what labels we use for different kinds of filters (the Gen2 coadds are definitely not consistent), but the good news is that the consistent thing is also easier here.

Anyhow, I think you just want to use `band` values in all of the plots and column names (at least for anything derived from coadd measurements), and drop the `filterMap` config option entirely.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,284 @@
...
+            return pipeBase.Struct(qaTractTable_forced=allCats)
+        elif self.config.dfDataset == "meas":
+            return pipeBase.Struct(qaTractTable_unforced=allCats)
+        else:
+            return None
+
+    def writeMetadata(self, dataRefList):
+        """No metadata to write.
+        """
+        pass
```
TallJimbo, line 284: No need for this `writeMetadata` is a Gen2-only method, and should only be present on `CmdLineTasks` (which some `PipelineTasks` _also_ are, but not this one).

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,284 @@
...
+        allCats["id"] = allCats.index
+        # Reindex to match sorting of objTable_tract tables
+        inputObjectTable = inputObjectTable.get()
+        allCats = allCats.reindex(inputObjectTable.index)
+        if self.config.dfDataset == "forced_src":
+            return pipeBase.Struct(qaTractTable_forced=allCats)
+        elif self.config.dfDataset == "meas":
+            return pipeBase.Struct(qaTractTable_unforced=allCats)
+        else:
+            return None
```
TallJimbo, line 279: Doesn't seem like it should be possible to configure this class to do a lot of work and then throw it all away.  If it actually is possible, we should add a config validate check that makes it impossible.  After that (or if it already should be impossible), this can be an `assert False, "Should not be possible according to configuration checks."` or similar.

But this ties into my big picture comment, so it may be irrelevant after you've addressed this.

TallJimbo (commented):
I've added a few line comments, but my one big-picture comment is that I'd like to change how the `forced_src` vs. `meas` logic coexists in this task.

(A) If you want one task that can do either, but not both, I think it'd be better to split this into two or three task: one for each of meas and forced, and either a base task for the common connections, config options, and logic (in some utility method the derived tasks could call in `run`), or just some utility functions to hold that logic instead.

(B) If you want one task that can potentially do both at once, I'd remove the `config.dfDataset` config option in favor of `config.doWriteUnforced` and `config.doWriteForced` boolean options, and change the big `elif` to just another `if` (along with whatever other logic changes would be necessary to allow both of those blocks to coexist).

As for which of those two options is better, I'm not sure, but here's what I would use to weigh the options:

- (A) is probably going to lead to cleaner code; I think you'll find that the config options and logic can be simplified if you more fully separate the two code paths.  I would probably approach this by writing two completely separate tasks with a lot of duplication, and only then go back and look at what they actually have in common.
- (B) is probably a smaller change from what you have now. 
- If we expect to usually want both forced and unforced catalogs written _and_ we think it'll be more efficient to do the common logic (and common I/O), then (B) is better.
- If we might want to make just one of these most of the time, and then make the other one only if we see something we want to investigate, then (A) is better, as it makes it much more straightforward to generate them one-at-a-time.

sr525:
> I've added a few line comments, but my one big-picture comment is that I'd like to change how the `forced_src` vs. `meas` logic coexists in this task.
> 
> (A) If you want one task that can do either, but not both, I think it'd be better to split this into two or three task: one for each of meas and forced, and either a base task for the common connections, config options, and logic (in some utility method the derived tasks could call in `run`), or just some utility functions to hold that logic instead.
> 
> (B) If you want one task that can potentially do both at once, I'd remove the `config.dfDataset` config option in favor of `config.doWriteUnforced` and `config.doWriteForced` boolean options, and change the big `elif` to just another `if` (along with whatever other logic changes would be necessary to allow both of those blocks to coexist).
> 
> As for which of those two options is better, I'm not sure, but here's what I would use to weigh the options:
> 
> * (A) is probably going to lead to cleaner code; I think you'll find that the config options and logic can be simplified if you more fully separate the two code paths.  I would probably approach this by writing two completely separate tasks with a lot of duplication, and only then go back and look at what they actually have in common.
> * (B) is probably a smaller change from what you have now.
> * If we expect to usually want both forced and unforced catalogs written _and_ we think it'll be more efficient to do the common logic (and common I/O), then (B) is better.
> * If we might want to make just one of these most of the time, and then make the other one only if we see something we want to investigate, then (A) is better, as it makes it much more straightforward to generate them one-at-a-time.

Can I add a vote for A? I think it fits more with the style I was aiming for for the package. Not claiming I've achieved that but it's good to have an aim.

laurenam:
'course you can!  A it is...at least I've given it a first attempt!  I've created a `BaseMakeQaTractTablesTask` from which the new `MakeForcedQaTractTablesTask` and `MakeUnforcedQaTractTablesTask`s inherit all the common logic.   I'm not sure if my use of the `dfDataset` config to control the behavior within the shared `run()` method is good practice, but it way my best effort towards minimizing (eliminating!) code duplication.

Also, for now, do ignore the `__init__.py hack...` commit.  I'm not sure what's going on, but without it, my new module is no longer being "found" (it’s now only finding the version module).  I fear there may be something amiss with my local environment...I've asked for some advice on #software-dev.

You can see the new `pipeliine`s here and in `obs_subaru` that default to run both tasks.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,326 @@
...
+class BaseMakeQaTractTablesTaskConfig(pipeBase.PipelineTaskConfig,
+                                      pipelineConnections=BaseMakeQaTractTablesTaskConnections):
+    filterMap = pexConfig.DictField(
+        keytype=str, itemtype=str,
+        default={},
+        doc=("Dictionary mapping canonical band to the camera's physical filter.  This is currently "
+             "required for {coaddName}Coadd_obj that are indexed on the latter as the dataId supplied "
+             "will only contain the former (this will likely become unnecessary once the postprocess "
+             "scripts are updated to index on band).")
+    )
```
TallJimbo, line 62: Should not need this config anymore; the `band` value you get from the data ID is the only filter value we want to use in the downstream plots.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,326 @@
...
+    dfDataset = pexConfig.ChoiceField(
+        dtype=str,
+        default="forced",
+        allowed={
+            "forced": "Measurement values come from the forced_src catalogs",
+            "unforced": "Measurement values come from the meas catalogs",
+        },
+        doc="Type of measurement values of interest, either forced or unforced."
+    )
```
TallJimbo, line 71: Shouldn't need this option either; it's effectively fixed to a constant by the subclass.

We can at least just make it a class-level constant in the subclass tasks, but I bet we can do better; I'll see if I can make a more concrete suggestion when we get down to that code.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,326 @@
...
+        name="qaTractTable_forced",
+        dimensions=("tract", "band", "skymap"),
+    )
+
+
+class MakeForcedQaTractTablesTaskConfig(BaseMakeQaTractTablesTaskConfig,
+                                        pipelineConnections=MakeForcedQaTractTablesTaskConnections):
+    def setDefaults(self):
+        BaseMakeQaTractTablesTaskConfig.setDefaults(self)
+        self.dfDataset = "forced"
```
TallJimbo, line 281: Shouldn't need this method anymore after you remove the config option.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,326 @@
...
+    # We want the following to come from the *_meas catalogs as they reflect
+    # what happened in SFP calibration.
+    columnsToCopyFromMeas = pexConfig.ListField(
+        dtype=str,
+        default=["calib_", ],
+        doc=("List of string \"prefixes\" to identify the columns to copy.  All columns with names "
+             "that start with one of these strings will be copied from the *_meas catalogs into the "
+             "*_forced_src catalogs UNLESS the full column name contains one of the strings listed "
+             "in the notInColumnStrList config.")
+    )
```
TallJimbo, line 81: This config option and the one below seem to only be used by the `MakeForcedQaTractTablesTask` subclass, not the `MakeUnforced...` variant.  They should be moved to the config class for just that subclass.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,326 @@
...
+                if measColsToLoadList is None:
+                    measColsToLoadList = [col for col in measCat.columns if
+                                          (col.startswith(tuple(self.config.columnsToCopyFromMeas))
+                                           and not any(s in col for s in self.config.notInColStrList))]
+                measCat = measCat[measColsToLoadList]
+                if not haveFilter:  # Fill in missing patch columns win NaNs
+                    cat = pd.DataFrame().reindex_like(cat)
+                    refCat = pd.DataFrame().reindex_like(refCat)
+                    measCat = pd.DataFrame().reindex_like(measCat)
+                cat = pd.concat([cat, refCat, measCat], axis=1)
```
TallJimbo, line 238: How about we move the code in this `if` block into a method that returns `cat` and takes whatever of the local variables it uses as arguments?  that can be an abstract method that raises `NotImplementedError` in the base class, have the above code in `MakeForced...` and code below in `MakeUnforced...`.

There might be a lot of arguments at first, and maybe more than just `cat` returned, so this may not be the final form, but I bet it'll be closer, and it'll make it clear what step to take next.

TallJimbo (commented):
I've left a few more restructuring comments.

I'd recommend going through and removing the `filterMap` config option and its usage first, and then do the restructuring into separate methods after that, just to reduce how much code is in motion in the latter step (I'd also recommend reading all line comments before acting on any of them - that will also make this sentence make a lot more sense).

laurenam:
Ok, I _think_ I've got things restructured as per you suggestions @TallJimbo, but do let me know if/where I have missed the mark!  One thing I waffled on was whether the new `@abstractmethod` `collateTable()` should be prepended with an underscore (I'm not very familiar with the criteria for designating a method as "private").  

I've added `TODO`s to remove the `filterMap` config song & dance once DM-28479 lands, but let me know if it's better to hold off merging this until that gets resolved.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,353 @@
...
+    @abstractmethod
+    def collateTable(self, inputCat, activeFilterName, haveFilter):
+        """Collect and aggregate catalog columns required for QA purposes.
+
+        Parameters
+        ----------
+        inputCat : `pandas.core.frame.DataFrame`
+            The input _obj table from which to collect the relevant data
+            columns.
+        activeFilterName : `int`
```
TallJimbo, line 130: `int` doesn't seem compatible with this being a name.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,353 @@
...
+
+class MakeForcedQaTractTablesTaskConnections(BaseMakeQaTractTablesTaskConnections,
+                                             defaultTemplates={"coaddName": "deep"},
+                                             dimensions=("tract", "skymap")):
+    qaTractTable = cT.Output(
+        doc="The collated catalog of forced measurements with additional columns for QA added",
+        storageClass="DataFrame",
+        name="qaTractTable_forced",
+        dimensions=("tract", "band", "skymap"),
+    )
```
TallJimbo, line 234: I've been wondering whether to recommend moving the output connection to the base class, and I think the fact that the base class's `run` clearly expects it to exist settles it.

More specifically, I think we should:
 - move the `qaTractTable = cT.Output(...)` from the `MakeForced` connections class to the base one;
 - in `MakeUnforcedQaTractTablesTaskConfig`, add a `setDefaults` with `self.connections.qaTractTable = "qaTractTable_unforced"`.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,353 @@
...
+    Subclasses must implement their own collateTable() method to collect the
+    relevant columns from the input tables.
+    """
+
+    ConfigClass = BaseMakeQaTractTablesTaskConfig
+    _DefaultName = "baseMakeQaTractTablesTask"
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        inputs["band"] = butlerQC.quantum.outputs["qaTractTable_" + self.MEASUREMENT_TYPE][0].dataId["band"]
```
TallJimbo, line 115:
```suggestion
        inputs["band"] = quantum.dataId["band"]
```

This is not just simpler; it's also strongly discouraged to use the actual dataset type names (like "qaTractTable_unforced") instead of the internal names (i.e. just "qaTractTable") inside a `PipelineTask`, as that prevents users from reconfiguring the task to write to or read from a different dataset type later.

It also lets you remove the `MEASUREMENT_TYPE` attributes from the derived classes, as this is the only place they now seem to be used.

TallJimbo (commented):
One more round of small comments.  Looking good!

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,353 @@
...
+        """Implementation of BaseMakeQaTractTablesTask.collateTable for forced
+        measurements.
+        """
+        colsToLoadList = None
+        refColsToLoadList = None
+        measColsToLoadList = None
+
+        measCat = inputCat["meas"][activeFilterName]
+        forcedCat = inputCat["forced_src"][activeFilterName]
+        if colsToLoadList is None:
```
sr525, line 286: If you've just set this to None how is it ever not None?

laurenam: Arghh...good point!  I guess I need to move the setting to `None` to `run` and pass them in (here and below in the unforced’s version) .  Thanks for catching this!

python/lsst/analysis/drp/makeQaTractTables.py
laurenam: Heh...I was kinda waiting for you to flag this one.  The problem with your suggestion (and why I didn't do it in the first place) is that this may be a rare case where there is no `band` dimension in my input connections since neither the `_obj` or `_objTable_tract` inputs have `band` as part of their `dataId`.  So, for example, the `dataId` I get from the above is:
```
print(butlerQC.quantum.dataId)
{skymap: 'hsc_rings_v1', tract: 9813}
```
However, inspired by your kick (along with the `cT.Output` shuffle), I looked further into the `outputRefs` object, and I think I've simplified this down to:
```
inputs["band"] = outputRefs.qaTractTable.dataId["band"]
```
which I hope doesn't fall under the "strongly discouraged" category!

laurenam:
Ready for another round of comments.  Again, I think I've implemented all the suggestions from you both.  I'm not entirely sure if my setting of the class attributes from within a class method is "best practice", but it's the only way I could think of to make sure the derivation of the desired-based-on-existing columns happen only once.  Along the way, I think I've sped things up a fair bit by making use of the "load only these columns" feature of DataFrames (it took me a minute to figure out the correct nested-dict syntax...I had almost convinced myself it wasn't possible for a second there!).  I foolishly failed to note the timing of the previous implementation...but by "feel", the script does now run quite a bit faster...
And, for the column deriving, I added a new function to call each time (it's small, but prolly better in the context of duplication).  Always open to (and most grateful for!) any further suggestions.

python/lsst/analysis/drp/makeQaTractTables.py
TallJimbo: Glad you caught my mistake, and your fix is exactly the right one!  (I was actually going to suggest that first, but I figured `quantum.dataId` would be even simpler.)

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,418 @@
...
+
+    See BaseMakeQaTractTablesTask for full description.
+    """
+
+    ConfigClass = MakeForcedQaTractTablesTaskConfig
+    _DefaultName = "makeForcedQaTractTablesTask"
+
+    colsToLoadList = None
+    refColsToLoadList = None
+    measColsToLoadList = None
```
TallJimbo, line 281: Your instincts that said we might not want to use class attributes like this were unfortunately correct, and I don't see a super easy fix; logic that basically says "do something special the first time through a loop" is pretty much always hard clean up.

This "I need to read one catalog in order to get the schema" problem is actually exactly why we have a practice of writing separate `_schema` datasets for all of our afw table catalogs, and I think this is a pretty clear sign that we need to be doing the same for our parquet tables.  That would let us load the schemas in `__init__`, compute the columns there, and make these variables frozen instance attributes.  That actually lets us chain schema-definition through multiple tasks - we can compute the final schema for some downstream task before actually running data through the first ones.

We won't get that done on this ticket, though (I'll open a new one for it).  Instead, let's use a pair of small helper class; something like this:

```
class ForcedQaLoadList:
     def __init__(self, inputCatCols, activeFilterName):
        self.forcedCols = ...  # compute cols to load from forced
        self.refCols = ... # compute cols to load from ref
        self.measCols = ... # compute cols to load from meas

class UnforcedQaLoadList:
    def __init__(self, inputCatCols, activeFilterName):
        self.cols = ...  # compute cols to load from meas
```
and a new abstract method:
```
@abstractmethod
def makeLoadList(self, inputCatCols, activeFilterName):
    """Make a "load list" instance appropriate for this derived class.
        ...
    """
        raise NotImplementedError()
```
The derived classes would implement that to just construct one of those helper classes.

The base class `run` method would then do something like this:
```
loadList = None
for inputCat in inputObjCats:
    ...
    if loadList is None:
        loadList = self.makeLoadList(inputCatCols, activeFilterName)
    cat = self.collateTable(..., loadList)
```
i.e. the base class would be responsible for figuring out when to make the load lists, but not what they contain, and it would then pass it into `collateTable` (which would need a new argument for it, but then _might_ be able to subtract some existing ones).

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,418 @@
...
+    def collateTable(self, inputCat, inputCatCols, activeFilterName, filterName, haveFilter):
+        """Implementation of BaseMakeQaTractTablesTask.collateTable for forced
+        measurements.
+        """
+        measTypeStr = "forced_src"
+        if self.colsToLoadList is None:
+            forcedColsList = [tup[2] for tup in inputCatCols if tup[0] == measTypeStr
+                              and tup[1] == activeFilterName]
+            self.colsToLoadList = deriveColumnsList(forcedColsList, self.config.baseColStrList,
+                                                    self.config.notInColStrList)
```
TallJimbo, line 292: It looks to me like the list comprehension that happens before a call to `deriveColumnsList` is always the same.  Would it make sense to move it inside `deriveColumnsList`, and make `deriveColumnsList` take `(measTypeStr, inputCatCols, activeFilterName)` instead of its current first argument?

TallJimbo (commented):
The problem of creating the load lists is a tricky one.  Hopefully the solution I've proposed in the line comments will work well enough for now.

python/lsst/analysis/drp/makeQaTractTables.py
```diff
@@ -0,0 +1,418 @@
...
+    to the persisted table, which has dataset type "qaTractTable_forced".
+
+    See BaseMakeQaTractTablesTask for full description.
+    """
+
+    ConfigClass = MakeForcedQaTractTablesTaskConfig
+    _DefaultName = "makeForcedQaTractTablesTask"
+
+    colsToLoadList = None
+    refColsToLoadList = None
```
sr525, line 280: Can we get rid of these with these changes?

python/lsst/analysis/drp/makeQaTractTables.py
laurenam: I also added an allowance for just a `list` of `str` being provided.

python/lsst/analysis/drp/makeQaTractTables.py
laurenam: First, a huge thanks for the master class you are giving me here...truly appreciated (and will hopefully pay in dividends for future reviewers of my code...if I can manage to retain any of the lessons learned here!)  So, I think I've got all your suggestions above implemented.  My current bad-practice speculation is my passing in the class `config`s to the new helper classes (seemed easier than accessing the individual config parameters and passing them in...but does it still make sense to have these `config`s defined where they are?)

TallJimbo: Passing in the `config` objects isn't that bad, and I think the config definitions themselves are fine now.  But I do think it'd be _slightly_ better to pass the config parameters separately instead; there are competing goals:
 - make code concise
 - don't pass a function _more_ than what it (or some interface it's trying to match) needs (in this case, all of the other config options those classes shouldn't care about).

I'd give the second of those more weight, generally, but this is the sort of thing I wouldn't have even noticed if you hadn't brought it up first.

TallJimbo (approved):

PR #7, opened 2021-07-12
DM-30923: Make first plots for coadd QA

sr525:


python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -151,3 +164,41 @@ def __call__(self, df, **kwargs):
...
+                      default=["g", "r", "i", "z", "y"])
+
+    def setDefaults(self):
+        super().setDefaults()
+        # This is missing one of the PA flags because they are not
+        # currently in the objectTablesbase_SdssCentroid_flag, is it
+        # redundant with xy_flag?
+        flagCols = ["PsfFlux_flag", "PixelFlags_saturatedCenter", "Extendedness_flag"]
+
+        filterColumns = ["xy_flag"] + [band + flag for flag in flagCols for band in self.bands]
```
ctslater, line 203: My understanding is that `setDefaults()` gets called before the user's configuration is applied, so in this function self.bands is always grizy and the config in the pipeline yaml never gets propagated to `self.selectWhenFalse`. (When I ran `coaddQAPlots.yaml` on a collection with only r and i-band data, all the plots were empty because the selector was seeing flags set on z and y band columns.)

sr525: Thank you for catching this! Some faffing later this is fixed.

ctslater: Excellent, I can confirm that the selector is now picking up the right bands. Plots of two-band data are still coming up empty for me though, and I think it's because the default `CoaddPlotFlagSelector` is still being applied (which requires grizy), in addition to the other selectors specified in the config yaml. I guess the config treats the additional selectors as an append rather than a replacement of the default?

(I didn't mean to pick such an edge case to start with, this is just the  `pipelines_check` data.)

sr525: Edge cases are good.

mrawls: Hey look, if I wait long enough to do a review, Colin finds broken things for me and Sophie fixes them! 😂

ctslater: Sweet, I have plots now! This is very exciting, and I will leave Meredith with the actual review (I'm not going to take all the work 🙂). There does seem to be a few places where the code throws un-graceful exceptions when no objects exist or when various selection arrays are empty, edge cases like that. Trying out a few less-than-fully-complete datasets like `pipelines_check` (or temporarily adding overly-stringent selectors) would show where additional guards would help.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -19,3 +21,20 @@ def setDefaults(self):
         super().setDefaults()
         self.colA.column = "iKronFlux"
         self.colB.column = "iPsFlux"
+
+
+class MagDiff(MultiColumnAction):
+    """Calculate the difference between two magntiudes"""
```
mrawls, line 27:
```suggestion
    """Calculate the difference between two magnitudes"""
```

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -19,3 +21,20 @@ def setDefaults(self):
...
+
+    col1 = Field(doc="Column to subtract from", dtype=str)
+    col2 = Field(doc="Column to subtract", dtype=str)
+
+    @property
+    def columns(self):
+        return (self.col1, self.col2)
+
+    def __call__(self, df):
```
mrawls, line 36: This has assumptions of units built into it. Can you please use astropy quantities and be more explicit about what units you expect the columns to have? Logging something like "subtracting col2 from col1 assuming flux units of X" might be worthwhile, for example.

python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -28,46 +28,47 @@ def __call__(self, df, **kwargs):
         """Select on the given flags"""
```
mrawls, line 28: In this function, and all functions throughout, please document the input arguments and returned values. If you don't want to do this for all instances of `__call__` and that's the main method that does anything, you could probably put it in the class definition instead, I think that jives with the developer docs guide.

But in sum, if there's a list of arguments and/or a return statement, docstring time please!

python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -12,9 +12,9 @@ class FlagSelector(DataFrameAction):
     selectWhenFalse = ListField(doc="Names of the flag columns to select on when False",
                                 dtype=str,
                                 optional=False,
-                                default=["base_PixelFlags_flag"])
+                                default=[])
 
-    selectWhenTrue = ListField(doc="Names of the flag columns to select on when False",
+    selectWhenTrue = ListField(doc="Names of the flag columns to select on when True",
```
mrawls, line 17: Was this just wrong before, or did you change the functionality?

python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -151,3 +139,53 @@ def __call__(self, df, **kwargs):
         sourceType = np.zeros(len(df))
         sourceType[gals] = 2
         return sourceType
+
+
+class UnknownIdentifier(DataFrameAction):
+    """Identifies un classified objects from the dataFrame and marks them as a
+       9 in the added sourceType column"""
+
+    band = Field(doc="The band the object is to be classified as an unkown in.",
```
mrawls, line 148:
```suggestion
    band = Field(doc="The band the object is to be classified as unknown in.",
```

python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -151,3 +139,53 @@ def __call__(self, df, **kwargs):
         sourceType = np.zeros(len(df))
         sourceType[gals] = 2
         return sourceType
+
+
+class UnknownIdentifier(DataFrameAction):
+    """Identifies un classified objects from the dataFrame and marks them as a
+       9 in the added sourceType column"""
```
mrawls, line 146: Where does 9 come from? Do we have to use a numeric value, or could we use "UNKNOWN" or something that's easier to avoid misinterpreting?

python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -151,3 +139,53 @@ def __call__(self, df, **kwargs):
...
+    def __call__(self, df, **kwargs):
+        """Identifies sources classed as unknowns"""
+        unknowns = (df[self.band + "Extendedness"] == 9.0)
+        sourceType = np.zeros(len(df))
+        sourceType[unknowns] = 9
+        return sourceType
+
+
+class CoaddPlotFlagSelector(DataFrameAction):
+    """The flags to use for selecting sources for coadd QA"""
```
mrawls, line 165: Please be a lot more explicit about what QA means here. Convince me, the clueless user, these are the flags I probably want to plot. What if I want to configure this and use different ones instead?

python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -324,8 +316,19 @@ def scatterPlotWithTwoHists(self, catPlot, plotInfo, sumStats, yLims=False, xLim
...
+            unknowns = (catPlot["sourceType"] == 9)
+            toPlotList = [(catPlot.loc[unknowns, xCol].values, catPlot.loc[unknowns, yCol].values, "green")]
+        if np.any(catPlot["sourceType"] == 10):
+            toPlotList = [(catPlot[xCol].values, catPlot[yCol].values, "purple")]
 
         for (xs, ys, color) in toPlotList:
+            [xs1, xs25, xs50, xs75, xs95, xs97] = np.nanpercentile(xs, [1, 25, 50, 75, 95, 97])
+            xScale = (xs97 - xs1)/20.0  # This is ~5% of the data range
+            # 40 was used as the number of bins because it looked good,
+            # might need to be changed in the future
```
mrawls, line 329: Can this be made configurable?

python/lsst/analysis/drp/skyPlot.py
```diff
@@ -262,11 +270,39 @@ def skyPlot(self, catPlot, plotInfo, sumStats):
...
+                            + "{:0.2f}\n".format(statAllMad) + r"n$_{points}$: "
+                            + "{}".format(len(catPlot)))
+            bbox = dict(facecolor="green", alpha=0.2, edgecolor="none")
+            ax.text(0.8, 0.91, allStatsText, transform=fig.transFigure, fontsize=8, bbox=bbox)
+
         toPlotList = []
         if np.any(catPlot["sourceType"] == 1):
             toPlotList.append((xsStars, ysStars, colorValsStars, "winter_r", "Stars"))
-        elif np.any(catPlot["sourceType"] == 2):
+        if np.any(catPlot["sourceType"] == 2):
```
mrawls, line 300: These could all be `elif`, and then you could log something at the end if there's anything left that's errantly corresponding to any other number.

pipelines/coaddQAPlots.yaml
```diff
@@ -0,0 +1,234 @@
+description: Make coadd plots for QA
```
mrawls, line 1: My main request for this lovely new yaml file is more documentation. Please explain what each plot is doing and pretend I have no idea (spoiler, I don't!).

pipelines/coaddQAPlots.yaml
```diff
@@ -0,0 +1,234 @@
...
+parameters:
+  ApFluxName: "iApFlux"
+  band: i
+tasks:
+  plot_CircAp12_sub_PS_meas_sky_stars:
+    class: lsst.analysis.drp.skyPlot.SkyPlotTask
```
mrawls, line 7: I'm a little surprised *ALL* of the plot tasks use the same class. I guess if you name it differently, you're allowed to use the same class multiple times in the same pipeline? Neat!

mrawls (commented):
I have not tried creating plots with the new pipelines, but overall I think this a nice way to automate some of the plots that were happening previously in pipe_analysis. Please see requests for more documentation etc., but otherwise basically good to go.

python/lsst/analysis/drp/dataSelectors.py
sr525: It was just wrong. Over enthusiastic copy and paste.

python/lsst/analysis/drp/dataSelectors.py
sr525: I will add more testing to the list of things to do once I have a few more plot types going.

python/lsst/analysis/drp/calcFunctors.py
sr525: I have added additional docs and used a config option for astropy units for the fluxes. Astropy doesn't have a jansky -> magnitude option nor does it have a millimag option so that doesn't really work for the outputs. The number of units that can be given is also relatively small as this action doesn't take a photocalib or any zeropoint information because it is expected to convert janskies to magnitudes. I will make this explicit in the docs.

mrawls: This is what @parejkoj recommended in a Community post a little while back, that I always refer to, if it helps
```
>>> import astropy.units as u
>>> (575*u.nJy).to(u.ABmag)
<Magnitude 24.50083039 mag(AB)>
>>> (0*u.ABmag).to_value(u.nJy)
3630780547701.003
```

parejkoj: Also `3631.0` is an approximation at a level that matters for our precision level. Just use astropy, as @mrawls says. (do the input/output data have units attached? If not, they really really should)

pipelines/coaddQAPlots.yaml
sr525: Currently all the plots are the same type but more of different types will be added with time. This is mostly just getting a skeleton in place while we develop more plots to add.

python/lsst/analysis/drp/calcFunctors.py
sr525: Somehow I missed the magAB option in the list of units. Thanks guys!

python/lsst/analysis/drp/skyPlot.py
sr525: These are all if deliberately so that it will plot all the different source types onto one plot if you want it to.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -19,3 +21,44 @@ def setDefaults(self):
...
+        flux1 = df[self.col1].values * u.Unit(self.fluxUnits1)
+        mag1 = flux1.to(u.ABmag)
+
+        flux2 = df[self.col2].values * u.Unit(self.fluxUnits2)
+        mag2 = flux2.to(u.ABmag)
+
+        return (mag1 - mag2)*1000.0
```
mrawls, line 64: Is multiplying by 1000 always desired here? The rest of the units looks great, thanks!

sr525: I think so, in all the plots I have made so far it has been. This can always be address in a later ticket if it is needed. Or a different action could be used.

PR #10, opened 2021-09-02
DM-28753: Create gen3 analysis_drp sky object sky plot

leeskelvin:


pipelines/coaddQAPlots.yaml
```diff
@@ -260,3 +260,18 @@ tasks:
         from lsst.analysis.drp.dataSelectors import (CoaddPlotFlagSelector, FlagSelector, GalaxyIdentifier,
         SnSelector, UnknownIdentifier)
         from lsst.analysis.drp.calcFunctors import MagDiff
+
+  plot_CircAp9_forced_skyObj_sky:
+    # Make a plot showing the 9-pixel radius circular aperture flux in units of nJy.
+    # This plot shows the on sky distribution of these values for sky objects.
+    class: lsst.analysis.drp.skyPlot.SkyPlotTask
+    config:
+      axisActions.zAction.column: "iAp9Flux"
```
cmsaunders, line 269: I see there is precedent from the gen2 version of this plot, but why use a 9-pixel aperture when all the others are 12? Is it too hard to find empty apertures if you use 12?

pipelines/coaddQAPlots.yaml
```diff
@@ -260,3 +260,18 @@ tasks:
...
+      connections.plotName: CircAp9_forced_skyObj_sky
+      selectorActions.flagSelector: FlagSelector
+      selectorActions.flagSelector.selectWhenTrue: ["merge_peak_sky"]
```
cmsaunders, line 272: All of the other plots have 
```
selectorActions.flagSelector: CoaddPlotFlagSelector
selectorActions.flagSelector.bands: ["I"]
```
Naively, I would assume you need the same here.

cmsaunders (commented):
This looks mostly fine. I just have a few questions: one below, plus a couple inline.

It looks like all the plots have 'Filter: N/A' in the information at the top. Is this fixable here or does it need to be addressed in `lsst.analysis.drp.skyPlot.SkyPlotTask`?

pipelines/coaddQAPlots.yaml
leeskelvin: Right, the 9-pixel aperture is the closest we have to an aperture that matches the "radius" of sky objects. Currently, sky objects are assigned to empty patches of sky where no other detected footprint exists within an 8-pixel radius. In fact, DM-23779 is tasked with increasing this from 8 to 9 pixels, to allow for our 9 pixel aperture to exactly match sky object apertures in future analyses.

pipelines/coaddQAPlots.yaml
leeskelvin: These parameters are set as default, so not explicitly required here. However, I agree that explicitly naming them may help future-proof against any updates to the defaults. Will edit, test and push later today.

leeskelvin:
Thanks for the review Clare. With regards the `Filter: N/A` issue, yes, this is a known issue in `lsst.analysis.drp.skyPlot.SkyPlotTask` that predates this ticket. It will be addressed in a future ticket.

pipelines/coaddQAPlots.yaml
leeskelvin: Having tested these edits, I've reverted back. Adding the selector actions present in all other plots causes the pipetask run to crash here. This is because we require a `FlagSelector` action to select explicitly using the flag column `merge_peak_sky`. This logic is different than the other plots in this pipeline. The `FlagSelector` action does not have a `bands` parameter and, therefore, it cannot be specified here. In which case, I think the ticket as-stands is the correct way to go.

cmsaunders (approved):

PR #12, opened 2022-01-21
Tickets/DM-32508

sr525:


python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -320,7 +360,7 @@ def __call__(self, df, **kwargs):
         """
 
         result = None
-        flagCols = ["_psfFlux_flag", "_pixelFlags_saturatedCenter", "_extendedness_flag"]
+        flagCols = ["psfFlux_flag", "pixelFlags_saturatedCenter", "extendedness_flag"]
         filterColumns = ["xy_flag"]
         filterColumns += [band + flag for flag in flagCols for band in self.bands]
```
mrawls, line 365: Strongly suggest doing the `_` string parsing here, and not defining anything with extra underscores (bands or flags). So it would be `band + '_' + flag for ...`

This happens in a few places, be sure to catch all of them.

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -81,6 +84,40 @@ def generateSummaryStats(cat, colName, skymap, plotInfo):
     return patchInfoDict
 
 
+def generateSummaryStatsVisit(cat, colName, visitSummaryTable, plotInfo):
+    """Generate a summary statistic in each patch
```
mrawls, line 88: Do you actually mean patch here, or do you mean visit or detector?

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -67,3 +68,69 @@ def __call__(self, df):
         if self.returnMillimags:
             magDiff = magDiff*1000.0
         return magDiff
+
+
+class CalcE1(MultiColumnAction):
```
mrawls, line 73: Class docstrings here and throughout, please

pipelines/visitQAEllipPlots.yaml
```diff
@@ -0,0 +1,364 @@
+description: Make visit level ellipticity plots for QA
```
mrawls, line 1: A slightly extended description would be wonderful here (and for the other yaml), like if you run this on a visit, you will get 20 plots and they will be saved as gifs and automatically tweeted.

mrawls (commented):
This is really cool! Thanks for forcing me to learn how this pipeline-plot-workflow-thing ought to work. I don't understand how all the selectors work very well, and I haven't gone through the particularities of each plot line-by-line since they (mostly) look OK, my random Slack ramblings notwithstanding. Most of my comments fall under "do flags better" and/or "what is this plot anyway." Hopefully straightforward.

python/lsst/analysis/drp/skyPlotVisit.py
```diff
@@ -0,0 +1,154 @@
...
+        runName : `str`
+            The name of the collection that the plot is written out to.
+        skymap : `lsst.skymap`
+            The skymap used to define the patch boundaries.
+        tableName : `str`
+            The type of table used to make the plot.
+
+        Returns
+        -------
+        `pipeBase.Struct` containing:
```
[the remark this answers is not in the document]
sr525: Returning a struct is standard and more things might be added to it later.

python/lsst/analysis/drp/skyPlotVisit.py
```diff
@@ -0,0 +1,154 @@
...
+        self.statisticSelectorActions.statSelector.threshold = 100
+
+
+class SkyPlotVisitTask(SkyPlotTask):
+
+    ConfigClass = SkyPlotVisitTaskConfig
+    _DefaultName = "skyPlotVisitTask"
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        # Docs inherited from base class
```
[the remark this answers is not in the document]
sr525: I can delete these if you want but I thought it was a fairly standard way of explaining why there are no docstrings.

python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -258,12 +258,52 @@ def __call__(self, df, **kwargs):
...
 
+class VisitPlotFlagSelector(DataFrameAction):
+
+    @property
+    def columns(self):
+        flagCols = ["psfFlux_flag", "pixelFlags_saturatedCenter", "extendedness_flag", "centroid_flag"]
+        yield from flagCols
+
+    def __call__(self, df, **kwargs):
+        """The flags to use for selecting sources for visit QA
```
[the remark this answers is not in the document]
sr525: I thought this was implied but I have clarified in the notes section.

python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -258,12 +258,52 @@ def __call__(self, df, **kwargs):
...
+        -----
+        These flags are taken from pipe_analysis and are considered to
+        be the standard flags for general QA plots. Some of the plots
+        will require a different set of flags, or additional ones on
+        top of the ones specified here. These should be specifed in
+        an additional selector rather than adding to this one.
+        """
+
+        result = None
+        flagCols = ["psfFlux_flag", "pixelFlags_saturatedCenter", "extendedness_flag", "centroid_flag"]
```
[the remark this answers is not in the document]
sr525: There is a selector that takes flag names as a config but it was decided to have selectors that stayed the same so that people could be sure they were reusing the same things when making plots.

python/lsst/analysis/drp/plotUtils.py
sr525: It works on either patches or detectors, I have clarified this.

python/lsst/analysis/drp/skyPlotVisit.py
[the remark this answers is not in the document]
mrawls: I'll concede this point for now, but I'm thinking about how to make these plotting utilities useful in a notebook, and returning a Struct is kindof the opposite of ideal.

python/lsst/analysis/drp/skyPlotVisit.py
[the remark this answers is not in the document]
mrawls: Fair enough 🙂

python/lsst/analysis/drp/dataSelectors.py
mrawls: I understand why you did it, but I'm concerned it will introduce bugs when clueless people do things like add a band and don't magically know that band names in this context should have trailing underscores. I think adding `+ '_'` each time you build a column list is a lot less inconvenient.

python/lsst/analysis/drp/dataSelectors.py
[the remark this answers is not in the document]
mrawls: The notes is helpful, but still missing the most basic of statements like "These flags are considered to be the standard flags **indicating a source is bad** for general QA plots." Sometimes I get confused and wonder if a list of flags means that sources with those flags are good.

python/lsst/analysis/drp/dataSelectors.py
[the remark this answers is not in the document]
mrawls: Hmm. I'm looking at line 271 in this same class, though, under the property `def columns(self)` you use repeatedly elsewhere, and not understanding why you can't use that here instead of redefining `flagCols` from scratch to match — it's the same four flags. Is this a special case for some reason?

python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -94,7 +94,7 @@ def __call__(self, df, **kwargs):
         """
 
         result = None
-        flagCols = ["_psfFlux_flag", "_pixelFlags_saturatedCenter", "_extendedness_flag"]
+        flagCols = ["psfFlux_flag", "pixelFlags_saturatedCenter", "extendedness_flag"]
```
mrawls, line 97: I missed this one before, can you use the `flagCols` from the property `columns` here? I see the default columns listed are different... why is that? It's confusing to have a list of 4 default flag columns in one case and then override it with a different set of 3 defaults flags.

python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -320,7 +360,7 @@ def __call__(self, df, **kwargs):
...
```
mrawls, line 363: This one is a redefinition too 😭

pipelines/visitQAEllipPlots.yaml
```diff
@@ -0,0 +1,367 @@
+description: |
+    Make visit level ellipticity plots for QA, makes plots of the e1, e2 and size measurements
+    for the PSF shape measurements and the shape measurements in the shape slot, typically
+    CModel. Makes 18 plots that get saved as pngs.
+tasks:
+  plot_e1_scatter_visit:
```
mrawls, line 6: You have wonderful little summaries for all the individual plots in `visitQAPlots.yaml`, and now these ones are maybe feeling sad and left out.

mrawls:
Per slack DM, it's fine with to leave all the redefined `flagCols` for now, even though I don't like it, so long as you clarify that the `columns` methods throughout serve a different purpose in docstrings. e.g., "This is a convenience method for specifying a subset of flag columns to load from the catalog/dataframe. It does not affect which flag columns are plotted."

python/lsst/analysis/drp/dataSelectors.py
mrawls: Per discussion, some columns have `_` and some don't, how fun. We came up with a slightly less awful standardization scheme along the lines of `filterColumns += [band + '_' + flag if len(band) > 0 else band + flag for flag in flagCols for band in self.bands]`.

PR #15, opened 2022-03-07
DM-33963: Add resource-stats gathering task.

TallJimbo:


python/lsst/analysis/drp/gatherResourceStatistics.py
```diff
@@ -0,0 +1,119 @@
...
+                multiple=True,
+                deferLoad=True,
+            )
+            # No actual name overrides permitted (these connections aren't
+            # individually configurable, because they're dynamically added),
+            # but for now we need to populate these internal things manually
+            # because super().__init__ can only populate them with the usual
+            # static connections.
+            self._nameOverrides[label] = dataset_type_name
+            self._typeNameToVarName[dataset_type_name] = label
```
TallJimbo, line 63: @natelust , here's a first draft of an attempt to write a connections class with some dynamic inputs, driven by configuration.  I think the two lines above are the worse bits, but it'd certainly also be nice to have a clear way to set both `self.inputs` and `self.allConnections` at the same time consistently, and to make it safe to do that either before or after delegating to `super`.

Curious to get your thoughts on both what I've done given the constraints we have now, and how we might make it easier (but feel free to just go ahead and work on the latter if you have a clear idea of what you want to do).

python/lsst/analysis/drp/gatherResourceStatistics.py
```diff
@@ -0,0 +1,119 @@
...
+            "task or method's memory usage and runtime."
+        )
+    )
+
+    def __init__(self, *, config):
+        super().__init__(config=config)
+        # Inject one new input connection for each configured label.
+        for label in config.grouped_labels():
+            self.inputs.add(label)
+            dataset_type_name = f"{label}_metadata"
```
timj, line 49: There is a `TaskDef.metadataDatasetName` that forms the `label_metadata` name.  That's used by `ctrl_mpexec`. Should we have a class method on `TaskDef` that can be used rather than hard-coding it in here?

natelust (approved):
Consider adding type annotations, at least where convenient. Even if nothing checks it now, each bit added now potentially makes things easier in the future.

python/lsst/analysis/drp/gatherResourceStatistics.py
```diff
@@ -0,0 +1,533 @@
...
+                label=output_label,
+                klass=this_tasks_name,
+                config=[config_ir],
+            )
+            output_pipeline_ir.tasks[output_label] = task_ir
+            output_subset.subset.add(output_label)
+        return Pipeline.fromIR(output_pipeline_ir)
+
+    @classmethod
+    def make_pipeline_cli(cls, argv):
```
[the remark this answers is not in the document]
TallJimbo: I think they're sufficiently closely tied to the task that they belong here, and there's no new state that we might want a new class to hold.  These have actually been removed anyway, but there are now QuantumGraph-generating methods for which the same question and answer applies.

python/lsst/analysis/drp/gatherResourceStatistics.py
```diff
@@ -0,0 +1,533 @@
...
+
+    Returnsx
+    -------
+    name : `str`
+        Compact name for the set of dimensions.  Guaranteed to have only
+        alphanumeric characters and underscores (providing that is also true
+        of all dimension names), and for
+    """
+    terms = set(dimensions.required.names)
+    while True:
```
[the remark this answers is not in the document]
TallJimbo: This has now been removed, as it's unnecessary with the task now processing one input metadata dataset type per configuration.

natelust:
On the type annotations, I see what you are saying, I am happy for you to leave them out if you prefer. However, I am not quite sure I buy that argument, there is no more danger than the docstring being wrong, which is also something that is not checked or really enforced. At least with type annotations an author is more likely to updated them if an interface changes by virtue of being right next to the argument, and are useful when cross-checking the docstring.

python/lsst/analysis/drp/gatherResourceStatistics.py
```diff
@@ -172,19 +180,153 @@ class GatherResourceStatisticsTask(PipelineTask):
...
+            type=str,
+            default=None,
+            help="Data ID expression used when querying for input metadata datasets.",
+        )
+        args = parser.parse_args(argv)
+        base_dataset_type_filter = re.compile(r"\w+_metadata")
+        if not args.dataset_types:
+            input_dataset_types = base_dataset_type_filter
+        else:
+            input_dataset_types = [fnmatch.translate(expr) for expr in args.dataset_types]
```
timj: As a note, we wrote `lsst.daf.butler.core.utils.gloToRegex` to handle this in our command line tooling so you might want to use that here to ensure consistency (and it has the advantage that if you pass in a non-regex it returns back the string and not a regex).

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -0,0 +1,550 @@
...
+        dimensions=(),
+        doc=(
+            "Table that aggregates memory and CPU usage statistics from one "
+            "or more tasks. "
+            "This will have one row for each data ID, with columns for each "
+            "task or method's memory usage and runtime."
+        ),
+    )
+    input_metadata = ct.Input(
+        "PLACEHOLDER_metadata",  # Should always be overridden.
```
natelust, line 74: `PLACEHOLDER` in both places seems to imply these are linked together. If that is true, would using a template make sense here?

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -0,0 +1,550 @@
...
+        storageClass="TaskMetadata",
+        dimensions=(),  # Actually set in __init__, according to configuration.
+        doc="Metadata dataset for another task to gather resource usage from.",
+        multiple=True,
+        deferLoad=True,
+    )
+
+    def __init__(self, *, config):
+        super().__init__(config=config)
```
natelust, line 83: I think you should validate the names were actually overloaded here, else PLACEHOLDER gets registered into butler repos if people run `--register-dataset-types` Which I think runs before graph builder to ensure the types that are registered are known to the butler.

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -0,0 +1,550 @@
...
+        that generating a graph for each of those separately is many orders of
+        magnitude faster (and *not* doing is disastrously slow).  As a
+        short-term workaround, this method can be used to generate a
+        `QuantumGraph` for just this task much more efficiently from the
+        metadata datasets already present in a set of collections.
+        """
+        # Group input metadata datasets by dataset type.
+        metadata_refs_by_dataset_type = defaultdict(set)
+        for ref in metadata_refs:
+            metadata_refs_by_dataset_type[ref.datasetType].add(ref)
```
natelust, line 224: This is a small point, but we have been bitten by this MANY times in the past at unexpected points, I feel that all `defaultdict`s should be converted to a normal `dict` as soon as the defaultness is done being needed as a matter of practice.

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -0,0 +1,550 @@
...
+        # Iterate over those groups, creating a configuration of
+        # this task and quanta for each one.
+        quanta_by_task_def = {}
+        for input_metadata_dataset_type, metadata_refs in metadata_refs_by_dataset_type.items():
+            if not input_metadata_dataset_type.name.endswith("_metadata"):
+                continue
+            input_task_label = input_metadata_dataset_type.name[:-len("_metadata")]
```
natelust, line 231: consider doing something like:
```suggestion
            input_task_label = input_metadata_dataset_type.name.replace("_metadata",'.')
```
or if you are afraid there is a `_metadata` somewhere else in the name, possibly regular expression or something. Its not bad persay like this, but it always feels off and weird looking in code to see strings used as sequences (although they are)

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -0,0 +1,550 @@
...
+                outputs[log_dataset_type] = [DatasetRef(log_dataset_type, data_id)]
+            quantum = Quantum(
+                taskName=task_def.taskName,
+                taskClass=task_def.taskClass,
+                dataId=data_id,
+                inputs={input_metadata_dataset_type: list(metadata_refs)},
+                outputs=outputs,
+            )
+            quanta_by_task_def[task_def] = {quantum}
+        return QuantumGraph(quanta_by_task_def)
```
natelust, line 279: Do you think you want to check for cycles here, or subset to connected, and verify you have the expected number of independent chains (possibly not, I think you have just a bunch of independent quanta, but I wanted to raise it so it would be on your mind.)

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -0,0 +1,550 @@
...
+        """
+        dimensions = universe.extract(self.config.dimensions)
+        # Transform input list into a dict keyed by data ID.
+        handles_by_data_id = {}
+        for handle in input_metadata:
+            handles_by_data_id[handle.dataId] = handle
+        n_rows = len(handles_by_data_id)
+        # Create a dict of empty column arrays that we'll ultimately make into
+        # a table.
+        columns = {
```
natelust, line 372: why use this dict of numpy arrays vs a recarray

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -0,0 +1,550 @@
...
+
+        Parameters
+        ----------
+        quantum_metadata : `lsst.pipe.base.TaskMetadata`
+            The nested metadata associated with the label "quantum" inside a
+            PipelineTask's metadata.
+        handle : `lsst.daf.butler.DeferredDatasetHandle`
+            Butler handle for the metadata dataset; used to identify the
+            metadata in diagnostic messages only.
+        warned_about_metadata_version : `bool`
```
natelust, line 422: For some reason I thought warnings themselves could be configured to only be emitted once

natelust (approved):

python/lsst/analysis/drp/gatherResourceStatistics.py
```diff
@@ -0,0 +1,533 @@
...
+        ),
+        keytype=str,
+        itemtype=ExtractResourceUsageConfig,
+        optional=False,
+        default={},
+    )
+
+
+class GatherResourceStatisticsTask(PipelineTask):
+    """A `PipelineTask` that gathers resource usage statistics from task
```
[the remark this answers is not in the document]
TallJimbo: I think if there is a `from lsst.pipe.base import PipelineTask` then just `PipeTask` will use it, but if you import it another way you need to qualify it.  Another reason to like that import style!

python/lsst/analysis/drp/gatherResourceUsage.py
TallJimbo: Good idea; I always forget about the templates.

python/lsst/analysis/drp/gatherResourceUsage.py
TallJimbo: :+1: Dataset type registration always runs after GraphBuilder, but worth doing anyway.

python/lsst/analysis/drp/gatherResourceUsage.py
TallJimbo: I'll do the regex; I'm not really bothered by the indexing (I consider that different from `Sequence`-ness, though I'm not sure I can defend that distinction in my brain), but rolling it into the check in the previous couple of lines seems to make sense.

python/lsst/analysis/drp/gatherResourceUsage.py
TallJimbo: Should always be a bunch of independent quanta.

python/lsst/analysis/drp/gatherResourceUsage.py
TallJimbo: `pandas.DataFrame` is a transpose of `recarray` but pretty much identical to a dict of numpy arrays under the hood.  It may still do a copy when I convert it to a `DataFrame`, but at least it won't need to do a transpose.

python/lsst/analysis/drp/gatherResourceUsage.py
TallJimbo: If so I'm not familiar with it, and the message does include the data ID so the messages wouldn't be identical anyway.

python/lsst/analysis/drp/gatherResourceUsage.py
natelust: Is that true? I thought there was a situation the graph builder could get into if the types didn't exist yet (I know we fixed some bug related to that, but I thought there was additional undefined behavior by not registering first in some cases)

PR #16, opened 2022-03-07
DM-31751: Make final aesthetic changes

sr525:


python/lsst/analysis/drp/colorColorPlot.py
```diff
@@ -191,19 +209,22 @@ def colorColorPlot(self, catPlot, plotInfo):
         starCbAx = fig.add_axes([0.89, 0.11, 0.04, 0.75])
         plt.colorbar(starPoints, cax=starCbAx, extend="both")
         magLabel = self.config.axisLabels["z"]
-        galCbAx.text(0.6, 0.5, magLabel + ": Galaxies", color="k", rotation="vertical",
-                     transform=galCbAx.transAxes, ha="center", va="center", fontsize=10)
-        starCbAx.text(0.6, 0.5, magLabel + ": Stars", color="k", rotation="vertical",
-                      transform=starCbAx.transAxes, ha="center", va="center", fontsize=10)
+        galText = galCbAx.text(0.6, 0.5, magLabel + ": Galaxies", color="k", rotation="vertical",
```
leeskelvin, line 212: This is why all the colour bar axis labels are slightly lower than they should be. Every instance of this should be changed to:```
.text(0.5, 0.5
```

File: python/lsst/analysis/drp/colorColorPlot.py [hunk 2]
```diff
@@ -191,19 +209,22 @@ def colorColorPlot(self, catPlot, plotInfo):
...
+                               transform=galCbAx.transAxes, ha="center", va="center", fontsize=10)
+        galText.set_path_effects([pathEffects.Stroke(linewidth=3, foreground="w"), pathEffects.Normal()])
+        starText = starCbAx.text(0.6, 0.5, magLabel + ": Stars", color="k", rotation="vertical",
```
leeskelvin, line 215: `.text(0.5, 0.5`

python/lsst/analysis/drp/skyPlot.py
```diff
@@ -294,41 +329,78 @@ def skyPlot(self, catPlot, plotInfo, sumStats):
...
+                           vmax=vmax)
+
+            else:
+                plotOut = ax.scatter(xs, ys, c=colorVals, cmap=cmap, s=7, vmin=vmin, vmax=vmax)
+
             cax = fig.add_axes([0.87 + i*0.04, 0.11, 0.04, 0.77])
-            plt.colorbar(scatterOut, cax=cax, extend="both")
+            plt.colorbar(plotOut, cax=cax, extend="both")
             colorBarLabel = "{}: {}".format(self.config.axisLabels["z"], label)
             text = cax.text(0.6, 0.5, colorBarLabel, color="k", rotation="vertical", transform=cax.transAxes,
```
leeskelvin, line 405: `0.6, 0.5 --> 0.5, 0.5`

python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -523,15 +537,16 @@ def scatterPlotWithTwoHists(self, catPlot, plotInfo, sumStats, yLims=False, xLim
...
-        cax = divider.append_axes("left", size="14%")
+        cax = divider.append_axes("left", pad=0, size="14%")
         plt.colorbar(collection, cax=cax, orientation="vertical")
         cax.yaxis.set_ticks_position("left")
         for label in cax.yaxis.get_ticklabels():
-            label.set_bbox(dict(facecolor="w", ec="none", alpha=0.5))
-            label.set_fontsize(8)
+            label.set_bbox(dict(facecolor="w", ec="none", alpha=0.7))
+            label.set_fontsize(6)
         cax.text(0.6, 0.5, "Median Value", color="k", rotation="vertical", transform=cax.transAxes,
```
leeskelvin, line 550: `.text(0.6, 0.5 --> .text(0.5, 0.5`

python/lsst/analysis/drp/colorColorPlot.py
```diff
@@ -152,8 +166,12 @@ def colorColorPlot(self, catPlot, plotInfo):
                        self.config.connections.plotName, self.config.axisLabels["x"],
                        self.config.axisLabels["y"])))
 
-        fig = plt.figure()
-        ax = fig.add_axes([0.12, 0.11, 0.65, 0.75])
+        # Define a new colormap
+        newBlues = mkColormap(["paleturquoise", "midnightblue"])
+        newReds = mkColormap(["lemonchiffon", "firebrick"])
+
+        fig = plt.figure(dpi=200)
```
leeskelvin, line 173: I think this needs to be a higher `dpi`. On my screen, 200 DPI still looks a bit pixelated. Would recommend increasing to 300 DPI; 300 DPI is the standard print resolution for high resolution output

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -193,12 +206,7 @@ def colorColorFitPlot(self, catPlot, plotInfo, fitParams):
...
-        r, g, b = colors.colorConverter.to_rgb("C0")
-        r1, g1, b1 = colors.colorConverter.to_rgb("midnightblue")
-        colorDict = {"blue": ((0.0, b, b), (1.0, b1, b1)), "red": ((0.0, r, r), (1.0, r1, r1)),
-                     "green": ((0.0, g, g), (1.0, g1, g1))}
-        colorDict["alpha"] = ((0.0, 0.0, 0.0), (0.05, 0.3, 0.3), (0.5, 0.8, 0.8), (1.0, 1.0, 1.0))
-        newBlues = colors.LinearSegmentedColormap("newBlues", colorDict)
+        newBlues = mkColormap(["paleturquoise", "midnightblue"])
 
         # Make a figure with three panels
         fig = plt.figure()
```
leeskelvin, line 212: As with other comments in this review, I think you need to set `dpi=300` here too.

python/lsst/analysis/drp/histPlot.py
```diff
@@ -271,7 +284,7 @@ def histPlot(self, catPlot, plotInfo, sumStats):
                       f"panel 1 = {self.config.leftPanelLabels}; "
                       f"panel 2 = {self.config.rightPanelLabels} ")
 
-        fig = plt.figure()
+        fig = plt.figure(dpi=200)
```
leeskelvin, line 287: Strongly recommend `dpi=300`.

python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -261,7 +274,7 @@ def scatterPlotWithTwoHists(self, catPlot, plotInfo, sumStats, yLims=False, xLim
         self.log.info("Plotting {}: the values of {} on a scatter plot.".format(
                       self.config.connections.plotName, self.config.axisLabels["y"]))
 
-        fig = plt.figure()
+        fig = plt.figure(dpi=200)
```
leeskelvin, line 277: Strongly recommend `dpi=300`.

python/lsst/analysis/drp/skyPlot.py
```diff
@@ -221,9 +247,14 @@ def skyPlot(self, catPlot, plotInfo, sumStats):
         self.log.info("Plotting {}: the values of {} on a sky plot.".format(
                       self.config.connections.plotName, self.config.axisLabels["z"]))
 
-        fig = plt.figure()
+        fig = plt.figure(dpi=200)
```
leeskelvin, line 250: Strongly recommend `dpi=300`.

leeskelvin:
I can't comment on this in the code, because it's too far from any changes you've made, but you need to edit the z-axis label in `pipelines/coaddQAPlots.yaml` in the `plot_ri_gr_psf` task from:
```
"z": "iPsfMag (mag)"
```
to
```
"z": "i_psfMag (mag)"
```

leeskelvin:
PS - in the same file as above, it would help my own sanity if things we're laid out `x`, `y`, `z` wherever possible, i.e.:
```
axisActions.yAction.returnMillimags: False
axisActions.xAction.returnMillimags: False
```
to
```
axisActions.xAction.returnMillimags: False
axisActions.yAction.returnMillimags: False
```

leeskelvin:
General comment: All plots which currently reference 'PS' for the point source flux should be updated to reference 'PSF', in line with prior naming convention changes elsewhere in the stack. This happens, for example, in the `plot_CircAp12_sub_PS_meas_sky_stars` plot in `coaddQAPlots.yaml`.

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -295,7 +309,56 @@ def perpDistance(p1, p2, points):
...
+    xs : `np.array`
+        An array of the values to sort
+
+    Returns
+    -------
+    ids : `np.array`
+    """
+
+    med = np.median(xs)
+    dists = np.fabs(xs - med)
```
leeskelvin, line 362: There's essentially no good reason to use `np.fabs` over `np.abs`, as far as I can tell, and `np.abs` is more readable.

python/lsst/analysis/drp/skyPlot.py
```diff
@@ -294,41 +329,78 @@ def skyPlot(self, catPlot, plotInfo, sumStats):
...
+                maxDec = np.max(ys)
             med = np.median(colorVals)
             mad = sigmaMad(colorVals)
-            vmin = med - 3*mad
-            vmax = med + 3*mad
-            scatterOut = ax.scatter(xs, ys, c=colorVals, cmap=cmap, s=4.0, vmin=vmin, vmax=vmax)
+            vmin = med - 2*mad
+            vmax = med + 2*mad
+            if self.config.fixAroundZero:
+                scaleEnd = np.max([np.fabs(vmin), np.fabs(vmax)])
```
leeskelvin, line 378: Would recommend switching for `np.abs` here too.

python/lsst/analysis/drp/skyPlot.py
```diff
@@ -294,41 +329,78 @@ def skyPlot(self, catPlot, plotInfo, sumStats):
...
+                                    vmax=vmax)
+                # find the most extreme 15% of points, because the list
+                # is ordered by the distance from the median this is just
+                # the final 15% of points
+                extremes = int(np.floor((len(xs)/100))*85)
+                ax.scatter(xs[extremes:], ys[extremes:], c=colorVals[extremes:], s=s, cmap=cmap, vmin=vmin,
+                           vmax=vmax)
+
+            else:
+                plotOut = ax.scatter(xs, ys, c=colorVals, cmap=cmap, s=7, vmin=vmin, vmax=vmax)
```
leeskelvin, line 400: After much testing, I find this setup to work much better:
```
            if len(xs) > 5000:
                s = 500/(len(xs)**0.5)
                lw = (s**0.5) / 10
                plotOut = ax.imshow(binnedStats.T, cmap=cmap,
                                    extent=[xEdges[0], xEdges[-1], yEdges[-1], yEdges[0]], vmin=vmin,
                                    vmax=vmax)
                # find the most extreme 15% of points, because the list
                # is ordered by the distance from the median this is just
                # the final 15% of points
                extremes = int(np.floor((len(xs)/100))*85)
                ax.scatter(xs[extremes:], ys[extremes:], c=colorVals[extremes:], s=s, cmap=cmap,
                           vmin=vmin, vmax=vmax, edgecolor="white", linewidths=lw)

            else:
                plotOut = ax.scatter(xs, ys, c=colorVals, cmap=cmap, s=7, vmin=vmin, vmax=vmax,
                                     edgecolor="white", linewidths=0.2)
```

leeskelvin: It reduces the drop-off in size for large numbers of data points, as I found they were becoming too small to see for large volume plots. Furthermore, it adds a small white border around each point. I also reduced the threshold at which the transition from scatter-plot to hybrid scatter/histogram takes place from 10000 to 5000 points. For the record, I'm still not a huge fan of this hybrid plot, but, I realize that it had group consensus in discussions we've had before, so I'm happy to go with it here.

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -135,14 +142,21 @@ def addPlotInfo(fig, plotInfo):
...
+    bandsText = "Bands: " + plotInfo["bands"]
+    SNText = "S/N: " + str(plotInfo["SN"])
+    visitText = "Visit: " + str(plotInfo["visit"])
+
+    plt.text(0.01, 0.92, datasetType, fontsize=8, alpha=0.6, transform=fig.transFigure)
+    plt.text(0.01, 0.89, tableType, fontsize=8, alpha=0.6, transform=fig.transFigure)
+    plt.text(0.01, 0.86, tractText, fontsize=8, alpha=0.6, transform=fig.transFigure)
+    plt.text(0.01, 0.83, visitText, fontsize=8, alpha=0.6, transform=fig.transFigure)
+    plt.text(0.01, 0.80, bandsText, fontsize=8, alpha=0.6, transform=fig.transFigure)
+    plt.text(0.01, 0.77, SNText, fontsize=8, alpha=0.6, transform=fig.transFigure)
```
leeskelvin, line 159: I don't think you can get away with so many printed lines for this method. More often than not, the information overlaps the plotting region. From my testing, I think you can guarantee no more than 4 lines worth of information can be squeezed in at the top of the figure.

leeskelvin: I've rewritten the `addPlotInfo` method to look like this:
```
def addPlotInfo(fig, plotInfo):
    """Add useful information to the plot

    Parameters
    ----------
    fig : `matplotlib.figure.Figure`
    plotInfo : `dict`

    Returns
    -------
    fig : `matplotlib.figure.Figure`
    """

    # TO DO: figure out how to get this information
    photoCalibDataset = "None"
    astrometryDataset = "None"

    plotName = f"{plotInfo['plotName']}"
    datasetType = f"\nRun: {plotInfo['run']}"
    datasetsUsed = f"\nPhotoCalib: {photoCalibDataset}, Astrometry: {astrometryDataset}"
    tableType = f"\nTable: {plotInfo['tractTableType']}"
    if str(plotInfo["tract"]) != "N/A":
        tractText = f", Tract: {str(plotInfo['tract'])}"
    else:
        tractText = ""
    if str(plotInfo["visit"]) != "N/A":
        visitText = f", Visit: {str(plotInfo['visit'])}"
    else:
        visitText = ""
    bandsText = f", Bands: {''.join(plotInfo['bands'].split(' '))}"
    SNText = f", S/N: {str(plotInfo['SN'])}"

    infoText = f"{datasetType}{datasetsUsed}{tableType}{tractText}{visitText}{bandsText}{SNText}"

    plt.text(0.01, 0.99, plotName, fontsize=8, transform=fig.transFigure, ha='left', va='top')
    plt.text(0.01, 0.99, infoText, fontsize=8, transform=fig.transFigure, ha='left', va='top', alpha=0.6)

    return fig
```

leeskelvin: This has a number of advantages, including not having to specify the position of each new line in the printed information (using '\n' instead). Furthermore, I've added some logic that only prints one of Tract and Visit, as printing them both does not make sense.

leeskelvin:
General point regarding plot names. Plotting information is being duplicated, because the various plotting tasks prepend their plot type to the front of the plot name. For example, the sky object histogram is named: `histPlot_skyObject_hist`. Similarly, the PSF flux S/N meas sky plot is named `skyPlot_PSFluxSN_meas_all_sky`. 

Instead, the plot tasks in their respective `.yaml` files should be renamed from `plot_*` to `skyPlot_*` or `histPlot_*`, where appropriate, and the redundant `_sky`, `_hist`, etc, should be removed from the `.plotName`.

python/lsst/analysis/drp/skyPlot.py
```diff
@@ -233,14 +264,18 @@ def skyPlot(self, catPlot, plotInfo, sumStats):
...
-        ysStars = catPlot.loc[stars, yCol]
-        colorValsStars = catPlot.loc[stars, zCol]
+        colorValsStars = catPlot.loc[stars, zCol].values
+        ids = extremaSort(colorValsStars)
+        xsStars = catPlot.loc[stars, xCol].values[ids]
+        ysStars = catPlot.loc[stars, yCol].values[ids]
+        colorValsStars = colorValsStars[ids]
 
         # Calculate some statistics
         if np.any(catPlot["sourceType"] == 2):
```
leeskelvin, line 281: The `ax.text` for source type == 2 (galaxies) is:
```
ax.text(0.63, 0.91, ...
```
however, for all other types (stars/unknown) it is:
```
ax.text(0.8, 0.91, ...
```
These should be reconciled to be the same (i.e., changing the galaxies value to 0.63).

python/lsst/analysis/drp/skyPlot.py
```diff
@@ -294,41 +329,78 @@ def skyPlot(self, catPlot, plotInfo, sumStats):
...
+            vmax = med + 2*mad
+            if self.config.fixAroundZero:
+                scaleEnd = np.max([np.fabs(vmin), np.fabs(vmax)])
+                vmin = -1*scaleEnd
+                vmax = scaleEnd
+            nBins = 50
+            xBins = np.linspace(minRa, maxRa, nBins)
+            yBins = np.linspace(minDec, maxDec, nBins)
+            binnedStats, xEdges, yEdges, binNums = binned_statistic_2d(xs, ys, colorVals, statistic="median",
+                                                                       bins=(xBins, yBins))
```
leeskelvin, line 385: You're not setting `xBins` and `yBins` here, rather, you're setting bin edges. I've modified this section a little, which makes that clearer, and also improves the legibility of HSC plots by aligning bins with patches:
```
            nBins = 45
            xBinEdges = np.linspace(minRa, maxRa, nBins+1)
            yBinEdges = np.linspace(minDec, maxDec, nBins+1)
            binnedStats, xEdges, yEdges, binNums = binned_statistic_2d(xs, ys, colorVals, statistic="median",
                                                                       bins=(xBinEdges, yBinEdges))
```

leeskelvin: The corner tract/patch plot colour bar labels in the `scatterTwoHistPlot`'s overlap with the main body figure. They need to be adjusted.

leeskelvin: For consistency, it would be nice if the n_galaxies and n_stars number counts in the `colorColorPlot` were printed in the same shaded box format as in the `skyPlot` plots.

python/lsst/analysis/drp/skyPlot.py
sr525: This was here so that if stars and galaxies were plotted on the same plot both could be seen. I was thinking that I would change these to all be the same but then people have mentioned maybe wanting to plot both so I think I will leave it.

sr525: Actually I made it figure out if two types were being plotted and only move it over if needed.

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -135,14 +142,25 @@ def addPlotInfo(fig, plotInfo):
...
-    plt.text(0.02, 0.95, datasetType, fontsize=8, alpha=0.8, transform=fig.transFigure)
+    plt.text(0.01, 0.99, plotInfo["plotName"], fontsize=8, transform=fig.transFigure, ha="left", va="top")
+
+    run = plotInfo['run']
+    datasetsUsed = f"\nPhotoCalib: {photocalibDataset}, Astrometry: {astroDataset}"
+    tableType = f"\nTable: {plotInfo['tractTableType']}"
+
+    dataIdText = ""
+    if str(plotInfo["tract"]) != "N/A":
+        dataIdText += f"Tract: {plotInfo['tract']}"
```
leeskelvin, line 153: The way you have this set up, you need to change this line to:
```
dataIdText += f", Tract: {plotInfo['tract']}"
```
Otherwise, the printed info block doesn't have a comma-space between the table and the `Tract:`.

That said, wouldn't it be cleaner to add your required commas and whitespaces directly into the `infoText` constructor below instead?

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -135,14 +142,25 @@ def addPlotInfo(fig, plotInfo):
...
+    if str(plotInfo["visit"]) != "N/A":
+        if dataIdText != "":
+            dataIdText += " ,"
+        dataIdText += f"Visit: {plotInfo['visit']}"
```
leeskelvin, line 157: Also here too: 
```
dataIdText += f", Visit: {plotInfo['visit']}"
```

python/lsst/analysis/drp/plotUtils.py
```diff
@@ -135,14 +142,25 @@ def addPlotInfo(fig, plotInfo):
...
     astroDataset = "None"
 
-    plt.text(0.02, 0.98, "Run:" + plotInfo["run"], fontsize=8, alpha=0.8, transform=fig.transFigure)
-    datasetType = ("Datasets used: photocalib: " + photocalibDataset + ", astrometry: " + astroDataset)
-    tableType = "Table: " + plotInfo["tractTableType"]
-    dataIdText = "Tract: " + str(plotInfo["tract"]) + " , Filter: " + plotInfo["filter"]
-    plt.text(0.02, 0.95, datasetType, fontsize=8, alpha=0.8, transform=fig.transFigure)
+    plt.text(0.01, 0.99, plotInfo["plotName"], fontsize=8, transform=fig.transFigure, ha="left", va="top")
+
+    run = plotInfo['run']
```
leeskelvin, line 147: Do these need to be "" instead of ''?

PR #17, opened 2022-03-07
DM-33966: Create PSF size residual plots

arunkannawadi:


pipelines/visitQAEllipPlots.yaml
```diff
@@ -270,6 +270,29 @@ tasks:
         from lsst.analysis.drp.calcFunctors import CalcShapeSize
         from lsst.pipe.tasks.dataFrameActions import SubtractColumns
 
+  plot_shapeSizeResidsDiff_scatter_visit:
```
leeskelvin, line 273: Would `plot_shapeSizeResidDiff_scatter_visit` make more sense here?, i.e., not plural?

leeskelvin, line 273: Would `plot_shapeSizeResidDiff_scatter_visit` make more sense here?, i.e., not plural?

pipelines/visitQAEllipPlots.yaml
```diff
@@ -270,6 +270,29 @@ tasks:
...
+      axisActions.yAction.colB: CalcShapeSize
+      axisActions.yAction.colB.colXx: "ixxPSF"
+      axisActions.yAction.colB.colYy: "iyyPSF"
+      connections.plotName: shapeFractionDiff_sizePSF_scatter
+      selectorActions.flagSelector: VisitPlotFlagSelector
+      selectorActions.SnSelector: SnSelector
+      selectorActions.SnSelector.bands: [""]
+      selectorActions.SnSelector.fluxType: "psfFlux"
+      selectorActions.SnSelector.threshold: 10
+      axisLabels: {"x": "PSF Magnitude (mag)", "mag": "PSF Magnitude (mag)",
```
leeskelvin, line 289: Is the `mag` entry in this dict required, as it seems to be a repeat of `x`?

pipelines/visitQAEllipPlots.yaml
```diff
@@ -270,6 +270,29 @@ tasks:
...
+                   "y": "Fractional shape difference (ixx - ixxPSF)"}
```
leeskelvin, line 290: Just spoke with you about this label only referencing `ixx`, but realize that this will be addessed in DM-33971, so that's okay.

leeskelvin, line 290: Just spoke with you about this label only referencing `ixx`, but realize that this will be addessed in DM-33971, so that's okay.

pipelines/visitQAEllipPlots.yaml
arunkannawadi: I have changed the name of the action of `FractionalDifferenceColumns`. Perhaps, I could rename this to `plot_shapeSizeFractionalDiff_scatter_visit`?

pipelines/visitQAEllipPlots.yaml
```diff
@@ -270,6 +270,29 @@ tasks:
...
+  plot_shapeSizeResidsDiff_scatter_visit:
+    class: lsst.analysis.drp.scatterPlotVisit.ScatterPlotVisitTask
+    config:
+      axisActions.yAction: FractionalResidualColumns
+      axisActions.yAction.colA: CalcShapeSize
+      axisActions.yAction.colA.colXx: "ixx"
+      axisActions.yAction.colA.colYy: "iyy"
+      axisActions.yAction.colB: CalcShapeSize
+      axisActions.yAction.colB.colXx: "ixxPSF"
+      axisActions.yAction.colB.colYy: "iyyPSF"
```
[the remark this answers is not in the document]
leeskelvin: Fair enough! From our discussion yesterday, I wasn't sure if you also need `ixy`. If not, that's fine!

pipelines/visitQAEllipPlots.yaml
arunkannawadi: This follows from `axisLabels` in other plots. I don't understand why `mag` and `x` entries are required when the string appears exactly once in the plot. This will be addressed in DM-33971 as well.

arunkannawadi: On second thoughts, I think `mag` entry is associated with the stats that appears on the bottom right corner of the plot.

PR #18, opened 2022-03-11
DM-33957: Make red galaxy truth residual and pull plots

erykoff:


python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -143,3 +143,180 @@ def columns(self):
...
+        color1_flux2 = df[self.color1_flux2].values*u.Unit(self.color1_flux2_units)
+        color1_mag2 = color1_flux2.to(u.ABmag).value
+
+        color2_flux1 = df[self.color2_flux1].values*u.Unit(self.color2_flux1_units)
+        color2_mag1 = color2_flux1.to(u.ABmag).value
+
+        color2_flux2 = df[self.color2_flux2].values*u.Unit(self.color2_flux2_units)
+        color2_mag2 = color2_flux2.to(u.ABmag).value
+
+        color1 = color1_mag1 - color1_mag2
```
arunkannawadi, line 219: This is fine for now, but we should consider this as a technical debt and refactor it later to have a functor/action to calculate colors and use the `DifferenceColumns` action to compute color difference. I also didn't look at it in too much detail, but wonder if these can be achieved through the data frame actions [here](https://github.com/lsst/pipe_tasks/blob/13b7538eaf671293994b26784b4df6d834d07738/python/lsst/pipe/tasks/dataFrameActions/_actions.py#L60).

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -143,3 +143,180 @@ def columns(self):
...
+    @property
+    def columns(self):
+        cols = (self.color1_flux1,
+                self.color1_flux2,
+                self.color2_flux1,
+                self.color2_flux2,
+                self.color1_flux1_err,
+                self.color1_flux2_err)
+
+        if self.color2_flux1_err and self.color2_flux2_err:
```
arunkannawadi, line 282: I don't quite follow why the errors for `color2` have to be treated separately this way.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -143,3 +143,180 @@ def columns(self):
...
+
+        color2_flux2 = df[self.color2_flux2].values*u.Unit(self.color2_flux2_units)
+        color2_mag2 = color2_flux2.to(u.ABmag).value
+        if self.color2_flux2_err:
+            color2_mag2_err = k*df[self.color2_flux2_err].values/df[self.color2_flux2].values
+        else:
+            color2_mag2_err = 0.0
+
+        color1 = color1_mag1 - color1_mag2
+        err1_sq = color1_mag1_err**2. + color1_mag2_err**2.
```
arunkannawadi, line 314: You could consider `np.hypot` of all for columns instead of squaring and taking `sqrt` later.

python/lsst/analysis/drp/redGalaxyTruthAssociation.py
```diff
@@ -0,0 +1,193 @@
...
+class RedGalaxyTruthAssociationTask(pipeBase.PipelineTask):
+
+    ConfigClass = RedGalaxyTruthAssociationConfig
+    _DefaultName = "redGalaxyTruthAssociationTask"
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        input_ref_dict = butlerQC.get(inputRefs)
+
+        # Determine columns to read
+        obj_columns = set({'objectId', 'patch', 'coord_ra', 'coord_dec'})
```
arunkannawadi, line 92: This should be double quotes. There are a few single quotes throughout.

arunkannawadi:
Some of the commits can (may be should) be squashed. For e.g, the commit about converting astropy units to numpy arrays also fixes a typo bug.

arunkannawadi (approved):

python/lsst/analysis/drp/calcFunctors.py
TallJimbo: I'm curious why using a composite action might be considered better here than a simple subtraction - I am very new to actually using these, so there may be something important I am not aware of, but I worry that this is a case where maximizing use of something we wrote is automatically assumed to be good.

I think the actual criteria for using ConfigurableAction composition instead of writing a non-composite action should be something like this:

- Would there actually be any use cases for swapping out the nested action?
- If so, is it also important for those other use cases to share the parent action from a correctness standpoint (i.e. if we had two non-composite actions instead, and they got out of sync, is that bug)?

For cases where it's just that the nested action is complex enough that we want to encourage it to be reused, I think I'd rather see that complexity moved to other normal classes or functions that could be reused instead.

I'm thinking we might want to raise this at a team meeting; maybe try to brainstorm some conventions and best practices and get them written down.

arunkannawadi: My thought was that computing a color given a pair of bands would be a common thing to do, and hence we should have an action that would take in `"g"` and `"r"` say and give colors. I am okay with either subtracting the color after that to compute `ColorDiff` or use `DifferenceColumn` action.

erykoff: I think that this is definitely touching on something philosophical that we should discuss in team meeting.  I am not a fan of the nested actions, actually.  I much prefer programming in python than trying to turn yaml/pexConfig into a programming language.
The way that the plotting tools can be reused and configured to do different actions for each column is great.  But I don't see the problem in each column type being a particular block of python code.

python/lsst/analysis/drp/calcFunctors.py
erykoff: In this case, we're subtracting an observed color (with errors) from a true color (without errors).  I can foresee a future where we want to subtract an observed color from a separate observed color (both with errors).  But if you try to subtract (say) a true color from a true color, both without errors, then you divide by 0 and it won't work.  
If you prefer I could treat the color1 errors like this as well, and then just raise in the config validation if you configure such that neither color1 nor color2 has errors.

python/lsst/analysis/drp/calcFunctors.py
arunkannawadi: Yes, topic for a team meeting for sure!

python/lsst/analysis/drp/calcFunctors.py
arunkannawadi: I think that'd be good. Alternatively, a comment along this line would also be sufficient.

python/lsst/analysis/drp/calcFunctors.py
erykoff: Upon further review, `np.hypot` only works on two arrays/values, so it's back to square (root) one.

arunkannawadi: Fair enough. All looks good to me.

PR #19, opened 2022-03-14
DM-33961: Generalize histogram plotting task to n-panels

leeskelvin:


python/lsst/analysis/drp/histPlot.py
```diff
@@ -139,19 +107,18 @@ class HistPlotTask(pipeBase.PipelineTask):
...
         # Docs inherited from base class
-        columnNames = set(["patch"])
-        bands = []
-        for actionStruct in [self.config.leftPanelActions,
-                             self.config.rightPanelActions,
-                             self.config.selectorActions]:
-            for action in actionStruct:
-                for col in action.columns:
-                    columnNames.add(col)
+        selectorColumn = [x for x in [x for x in self.config.selectorActions][0].columns][0]
```
sr525, line 110: Can x be something more descriptive?

python/lsst/analysis/drp/histPlot.py
```diff
@@ -237,20 +199,20 @@ def run(self, catPlot, dataId, runName, skymap, tableName, bands, plotName):
         # Get useful information about the plot
         plotInfo = parsePlotInfo(dataId, runName, tableName, bands, plotName, SN)
         # Calculate the corners of the patches and some associated stats
-        sumStats = generateSummaryStats(plotDf, self.config.summaryStatsLabel, skymap, plotInfo)
+        sumStats = generateSummaryStats(datPlot, self.config.summaryStatsColumn, skymap, plotInfo)
```
sr525, line 202: Why did the name get changed to datPlot? It would be nice to have consistency across all the plot types.

python/lsst/analysis/drp/histPlot.py
```diff
@@ -274,94 +236,95 @@ def histPlot(self, catPlot, plotInfo, sumStats):
...
+
+        # determine gridspec divisions
+        summary_block = 40
+        panel_pad = 8
+        if len(panels) <= 2:
+            ncols = 1
+        else:
+            ncols = 2
+        nrows = int(np.ceil(len(panels) / ncols))
+        col_bounds = np.linspace(0+summary_block+panel_pad, 240, ncols+1, dtype=int)
```
sr525, line 259: Spaces needed.

sr525: Here and throughout, did you run PEP8?

python/lsst/analysis/drp/histPlot.py
```diff
@@ -274,94 +236,95 @@ def histPlot(self, catPlot, plotInfo, sumStats):
...
+                vUppers.append(pvals[1])
+            vLower = np.min(vLowers)
+            vUpper = np.max(vUppers)
+
+            # generate plot
+            ax = fig.add_subplot(gs[row_start:row_stop, col_start:col_stop])
+            for count, (hist, column) in enumerate(zip(hists, columns)):
+                isfinite = np.isfinite(datPlot[column])
+                datMed = np.nanmedian(datPlot[column][isfinite])
+                datMad = sigmaMad(datPlot[column][isfinite])
```
sr525, line 295: You should omit the NaNs from this.

python/lsst/analysis/drp/histPlot.py
```diff
@@ -274,94 +236,95 @@ def histPlot(self, catPlot, plotInfo, sumStats):
...
-        rightPanelAx.tick_params(labelsize=7)
-        yStep = 0.038*len(self.config.leftPanelActions.fieldNames) + 0.01
-        rightPanelAx.legend(loc="upper left", bbox_to_anchor=(0.00, 0.77-yStep),
-                            bbox_transform=fig.transFigure, fontsize=7, handleheight=1.5)
-        rightPanelAx.axvline(0, ls='--', lw=2, c='k')
-
-        # Corner plot of patches showing summary stat in each
-        axCorner = plt.gcf().add_subplot(gs[70:96, 3:21])
+        gs = gridspec.GridSpec(240, 240)
+
```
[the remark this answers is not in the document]
sr525: I think that doing it without gridspec would make it easier to read and involve less faffing.

python/lsst/analysis/drp/histPlot.py
```diff
@@ -274,94 +236,95 @@ def histPlot(self, catPlot, plotInfo, sumStats):
...
+            ncols = 1
+        else:
+            ncols = 2
+        nrows = int(np.ceil(len(panels) / ncols))
+        col_bounds = np.linspace(0+summary_block+panel_pad, 240, ncols+1, dtype=int)
+        row_bounds = np.linspace(0, 240, nrows+1, dtype=int)
+        col_starts, col_stops, row_starts, row_stops = [], [], [], []
+        for i in range(len(panels)):
+            col_starts.append(col_bounds[i%ncols]+panel_pad)
+            if (i == (len(panels)-1)) and (len(panels)%2 == 1):
```
sr525, line 264: Spaces

python/lsst/analysis/drp/histPlot.py
```diff
@@ -274,94 +236,95 @@ def histPlot(self, catPlot, plotInfo, sumStats):
...
+                col_stops.append(np.max(col_bounds)-panel_pad)
```
sr525, line 265: Spaces

python/lsst/analysis/drp/histPlot.py
```diff
@@ -274,94 +236,95 @@ def histPlot(self, catPlot, plotInfo, sumStats):
...
+            else:
+                col_stops.append(col_bounds[(i%ncols)+1]-panel_pad)
```
sr525, line 267: Spaces, I know you wanted a comment everytime there is an error but there are too many.

python/lsst/analysis/drp/histPlot.py
```diff
@@ -274,94 +236,95 @@ def histPlot(self, catPlot, plotInfo, sumStats):
...
+            ax.set_yscale(self.config.panels[panel].yscale)
+            ax.set_xlim(vLower, vUpper)
+            ylims = ax.get_ylim()
+            ax.set_ylim(ylims[0], ylims[1]*1.1)
+            ax.set_xlabel(self.config.panels[panel].label, labelpad=1)
+            ax.tick_params(labelsize=7)
+            # add legends
+            leg1 = ax.legend(loc="upper left", fontsize=7, frameon=False, handlelength=0, borderaxespad=0,
+                             handletextpad=0, handleheight=0.7, framealpha=0.4)
+            _ = plt.gca().add_artist(leg1)
```
sr525, line 310: Do you need a return thing here?

python/lsst/analysis/drp/histPlot.py
```diff
@@ -274,94 +236,95 @@ def histPlot(self, catPlot, plotInfo, sumStats):
...
+            for count, text in enumerate(leg1.get_texts()):
+                text.set_text(text.get_text().split(":::")[0])
+                text.set_color(plt.cm.tab10(count))
+            leg2 = ax.legend(loc="upper right", fontsize=7, frameon=False, handlelength=0, borderaxespad=0,
+                             handletextpad=0, handleheight=0.7, framealpha=0.4)
+            for count, text in enumerate(leg2.get_texts()):
+                text.set_text(text.get_text().split(":::")[1])
+                text.set_color(plt.cm.tab10(count))
+            renderer = fig.canvas.get_renderer()
+            tlens = [t.get_window_extent(renderer).width for t in leg2.get_texts()]
```
sr525, line 320: Some comments here, and in other places, would be useful to know what is going on.

python/lsst/analysis/drp/histPlot.py
leeskelvin: Changed this line to:```
selectorColumns = [col for selector in self.config.selectorActions for col in selector.columns]
```

File: python/lsst/analysis/drp/histPlot.py (hunk 2 above)
Comment (leeskelvin) [reply python/lsst/analysis/drp/histPlot.py:202]: Changed back to `plotDf`.

File: python/lsst/analysis/drp/histPlot.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (leeskelvin) [reply python/lsst/analysis/drp/histPlot.py:250]: Thanks for this comment Sophie. I played around with this for a long time, as initially, I agreed with you. In the end however, I discovered that doing what we want to do using subplots is actually more of a faff than using gridspec. For an N-panel figure, with a side panel axis reserved for the corner plot, the logic will always be a little complicated no matter if you use gridspec or subplots. If you don't object too strongly, and as this currently works as expected, I'll recommend keeping this as it is for now.

File: python/lsst/analysis/drp/histPlot.py (hunk 3 above)
Comment (leeskelvin) [reply python/lsst/analysis/drp/histPlot.py:259]: The entire `histPlot.py` script has been PEP8'ed. With the exception of `E501 line too long`, all PEP8 violations have been resolved.

File: python/lsst/analysis/drp/histPlot.py (hunk 4 above)
Comment (leeskelvin) [reply python/lsst/analysis/drp/histPlot.py:295]: The median values are now calculated further up the script, alongside the x-axis limits. The relevant median line now looks like this:
```
meds.append(np.nanmedian(hist_data))
```
I've removed the `[isfinite]` selection from the data selector, so use of `nanmedian` is required in case any non-finite values creep into the calculation.

File: python/lsst/analysis/drp/histPlot.py (hunk 9 above)
Comment (leeskelvin) [reply python/lsst/analysis/drp/histPlot.py:310]: The entire legend section has now been refactored to something much simpler, using `ax.text`.

File: python/lsst/analysis/drp/histPlot.py (hunk 10 above)
Comment (leeskelvin) [reply python/lsst/analysis/drp/histPlot.py:320]: This section has been simplified. Additionally, a number of in-line comments have been added explaining the various steps of data processing.

**Comment by leeskelvin** [comment]
Thanks for the above Sophie, much appreciated. In addition, you left me a couple of comments on Slack too:

> If you have generalised this to multiple panels and different things are being plotted on each then does the corner plot make sense anymore?

Good question, and the answer will sometimes certainly be: definitely not. I propose that a future ticket adds a toggleable True/False flag asking if the user wants a corner plot to be added or not. For the purposes of this ticket however, I propose to maintain the functionality of the current task and leave it in.

> the fontsize (for stats) could be smaller.

If I make it much smaller, it looks too small on the 1- and 2-panel plots. I could make this user-configurable however?

Thanks again. The whole thing has been pep-8'ed with `pycodestyle`, has passed the linter here on GH, and is currently running on Jenkins. Will post more on the Jira ticket.

**Comment by leeskelvin** [comment]
Oh, and I forgot the most important question:

> is there a way to add star/galaxy information?

Yes. I've now added a `selectors` property into the `HistPlotConfig` class which allows the user to specify a selector for each plotted histogram. The global selectors (`selectorActions`) still exist, and will still trim the data down across the entire plot. However, if the user wants to have unique selectors for each histogram, that is now possible. I'll attach an example of this to the Jira ticket, for reference. I've added the plotting task that produces this figure into the `coaddQAPlots.yaml` pipeline, for reference.

**Comment by leeskelvin** [comment]
PS, you also mentioned in passing that it would be nice to change the colour scheme. I can also make this a user-configurable option, if desirable?

File: python/lsst/analysis/drp/histPlot.py (hunk 4 above)
Comment (leeskelvin) [reply python/lsst/analysis/drp/histPlot.py]: Nevermind, I went back to using the `isfinite` logic. This section now looks like this:
```
isfinite = np.isfinite(hist_data)
meds.append(np.median(hist_data[isfinite]))
mads.append(sigmaMad(hist_data[isfinite]))
nums.append(np.sum(isfinite))
```

PR #25, opened 2022-03-16
DM-339933: Correct stellar locus plots for extinction

yalsayyad:


arunkannawadi:
Have same comments as Eli[ posted on Slack](https://lsstc.slack.com/archives/GU3E59F3N/p1647406804529649?thread_ts=1647406521.912789&cid=GU3E59F3N) but otherwise all looks good to me.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -71,6 +72,43 @@ def __call__(self, df):
...
+    @property
+    def columns(self):
+        return self.magDiff.columns + (self.ebv,)
+
+    def __call__(self, df):
+        diff = self.magDiff(df)
+        ebv = df[self.ebv].values
+        col1Band = self.magDiff.col1.split('_')[0]
+        col2Band = self.magDiff.col2.split('_')[0]
+        av1 = self.extinctionCoeffs[col1Band]
```
sr525, line 103: It might be better with a useful error message here in case someone tries to run it with a filter name not in the dictionary.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -71,6 +75,39 @@ def __call__(self, df):
         return magDiff
 
 
+class ExtinctionCorrectedMagDiff(DataFrameAction):
+    magDiff = ConfigurableActionField(doc="Action that returns a difference in magnitudes",
+                                          default=MagDiff, dtype=DataFrameAction)
+    ebv = Field(doc="E(B-V) Column", dtype=str, default="ebv")
+    extinctionCoeffs = DictField(
+        doc="Dictionary of extinction coefficients for conversion from E(B-V) to extinction, A_filter."
```
erykoff, line 83: This should be `A_band`. not `A_filter`.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -71,6 +75,39 @@ def __call__(self, df):
...
+        default=None)
+
+    @property
+    def columns(self):
+        return self.magDiff.columns + (self.ebv,)
+
+    def __call__(self, df):
+        diff = self.magDiff(df)
+        if not self.extinctionCoeffs:
+            _LOG.warn("No extinction Coefficients. Not applying extinction correction")
```
erykoff, line 95: This should be `_LOG.warning`, as `warn` is deprecated.

python/lsst/analysis/drp/calcFunctors.py
erykoff: Yes, either a more helpful error or (alternatively) log a warning as above and don't apply extinction corrections. (And below for `col2Band`).

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -71,6 +75,39 @@ def __call__(self, df):
...
+            return diff
+
+        ebv = df[self.ebv].values
+        col1Band = self.magDiff.col1.split('_')[0]
```
erykoff, line 99: Can we always guarantee that the band will be the first part of the split on `'_'` or does this need to be another configuration parameter?
(Also, I don't have an easy answer to this, but I was getting confused because I was reading `col` as `color` but here it means `column` and has been commented elsewhere we have some naming inconsistencies and technical debt, and is not going to be fixed on this PR.)

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -71,6 +75,39 @@ def __call__(self, df):
         return magDiff
 
 
+class ExtinctionCorrectedMagDiff(DataFrameAction):
+    magDiff = ConfigurableActionField(doc="Action that returns a difference in magnitudes",
+                                          default=MagDiff, dtype=DataFrameAction)
+    ebv = Field(doc="E(B-V) Column", dtype=str, default="ebv")
```
erykoff, line 81: I think this should be `colEbv` or something like that, because it's a column not a value.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -71,6 +75,39 @@ def __call__(self, df):
...
+        col1Band = self.magDiff.col1.split('_')[0]
+        col2Band = self.magDiff.col2.split('_')[0]
+        av1 = self.extinctionCoeffs[col1Band]
+        av2 = self.extinctionCoeffs[col2Band]
+        correction = (av1 - av2) * ebv
+
+        if self.magDiff.returnMillimags:
+            correction = correction*1000.0
+
+        return diff - correction*u.mag
```
erykoff, line 108: Upon further review, this is incorrect here and elsewhere (everywhere?) in `analysis_drp`, I fear.  The units should not be `u.mag` but `u.mmag` if we're using millimags.  This should be a separate ticket (@sr525 ).

erykoff (approved):
Some minor changes, and a follow-up ticket.

python/lsst/analysis/drp/calcFunctors.py
yalsayyad: Too late to change the naming convention of the Object Table now! 

But if you want the band to be optionally supplied per config parameters you got it.

python/lsst/analysis/drp/calcFunctors.py
yalsayyad: No prob. I changed MagDiff and this Action. I went to give your `ColorDiff` the same treatment but found that it was just a numpy array, so I'll be leaving the 1000* there. 
```    color_diff = color_diff.to(u.mmag)
AttributeError: 'numpy.ndarray' object has no attribute 'to'
```

PR #30, opened 2022-03-21
DM-34133: Add ScatterPlotWithTwoHistsTaskTestCase

taranu:


python/lsst/analysis/drp/scatterPlot.py
```diff
@@ -57,6 +57,35 @@ class ScatterPlotWithTwoHistsTaskConfig(pipeBase.PipelineTaskConfig,
         itemtype=str
     )
 
+    def get_requirements(self):
```
arunkannawadi, line 60: This should have an underscore in the beginning, since this is not meant to be a public API (or is it?). I don't have a good alternative, but `requirements` seems strange. `_get_bands_and_columns`, may be?

tests/test_scatterPlot.py
```diff
@@ -0,0 +1,131 @@
...
+        config.nonBandColumnPrefixes.append('refcat')
+        config.sourceSelectorActions.galaxySelector = GalaxyIdentifier
+        config.highSnStatisticSelectorActions.statSelector.threshold = 50
+        config.lowSnStatisticSelectorActions.statSelector.threshold = 20
+        self.task = ScatterPlotWithTwoHistsTask(config=config)
+
+        n = len(flux)
+        self.bands, columns = config.get_requirements()
+        data = {
+            'refcat_flux': flux,
```
arunkannawadi, line 76: There are quite a few single quotes in this file.

tests/test_scatterPlot.py
```diff
@@ -0,0 +1,131 @@
...
+                               runName='test',
+                               skymap=None,
+                               tableName='test',
+                               bands=self.bands,
+                               plotName='test')
+
+        filename_figure_tmp = os.path.join(self.testDir, "test_scatterPlot.png")
+        result.scatterPlot.savefig(filename_figure_tmp)
+        diff = compare_images(filename_figure_tmp, filename_figure_ref, 0)
+        self.assertIsNone(diff)
```
arunkannawadi, line 118: This is likely not going to give a helpful error message.

erykoff: This is the correct way to check/raise: https://github.com/LSSTDESC/skyproj/blob/main/tests/test_plotting.py#L47-L48

I'm sure the tolerance will have to be >0 because there are small differences on different platforms.

tests/test_scatterPlot.py
```diff
@@ -0,0 +1,131 @@
...
+        self.data = pd.DataFrame(data)
+
+    def tearDown(self):
+        if os.path.exists(self.testDir):
+            shutil.rmtree(self.testDir, True)
+        del self.bands
+        del self.data
+        del self.task
+
+    def test_ScatterPlotWithTwoHistsTask(self):
```
erykoff, line 106: You also need to protect plotting tests with the default matplotlibrc:
https://github.com/LSSTDESC/skyproj/blob/702ac340e46cccae37ad91194d017816bf596ceb/tests/test_plotting.py#L17

tests/test_scatterPlot.py
```diff
@@ -0,0 +1,131 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+import unittest
+import lsst.utils.tests
+
```
erykoff, line 25: For testing with matplotlib, you need this at the top: https://github.com/LSSTDESC/skyproj/blob/702ac340e46cccae37ad91194d017816bf596ceb/tests/test_plotting.py#L4

tests/test_scatterPlot.py
taranu: Well, it passed Jenkins with 0 tolerance.
compare_images returns a human-readable message on failure, and pytest prints this out if the assert fails, so what's the benefit of raising an exception in a test?

python/lsst/analysis/drp/scatterPlot.py
arunkannawadi: Sure, it doesn't change state, but it shows up as a method of the `config` in autocomplete or when used with `dir()`. If the user will never need to call this method, then best making it private. But up to you.

tests/test_scatterPlot.py
arunkannawadi: I triggered a failure, which is not hard to do - having repeatedly done that last week, and this was the error message I got.

```
AssertionError: 'Error: Image files did not match.\n  RMS Value: 11.985321431153329\n  Expected:  \n    /project/kannawad/repo/analysis_drp/tests/test_outputahus36lt/test_scatterPlot.png\n  Actual:    \n    /project/kannawad/repo/analysis_drp/tests/data/test_scatterPlot.png\n  Difference:\n    /project/kannawad/repo/analysis_drp/tests/data/test_scatterPlot-failed-diff.png\n  Tolerance: \n    0' is not None
```

While the message is a bit cryptic, I doubt there can be any more helpful message than having "Image files did not match".

arunkannawadi (approved):

tests/test_scatterPlot.py
erykoff: Did it print with the `\n`s like that?  Because if you do the matplotlib raise then it'll print the message correctly formatted.

arunkannawadi: It was verbatim, with the `\n`s like that.

image

taranu: Ok, raise it is then.

PR #31, opened 2022-04-06
tickets/DM-34349 Make scons tests depend on building pipelines

natelust:
Ensure that the pipelines have all been built prior to running any
tests.


ktlim (approved):
This looks correct.  I might expect this kind of logic to be in the top-level SConstruct rather than down here in SConscript, but I believe it should work here.  You might want to use double-quotes for consistency (and future black-ification).

PR #38, opened 2022-05-03
DM-34402: Allow specify the type of size and ellipticities in calcFunctors

arunkannawadi:


arunkannawadi:
The documentation-only changes are for the convenience of the reviewer and can be squashed with the previous commits.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -132,43 +132,79 @@ def __call__(self, df):
 
 
 class CalcE(MultiColumnAction):
```
plazas, line 134: Should this class `CalcE` and the others be listed in the `__all__` at the top of the file?

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -132,43 +132,79 @@ def __call__(self, df):
...
+    """Calculate a complex value representation of the ellipticity.
 
     The complex ellipticity is typically defined as
-    E = ((ixx - iyy) + 1j*(2*ixy))/(ixx + iyy) = |E|exp(i*2*theta).
-
-    For plotting purposes we might want to plot |E|*exp(i*theta).
-    If `halvePhaseAngle` config parameter is set to `True`, then
-    the returned quantity therefore corresponds to |E|*exp(i*theta)
-    """
+    E = |E|exp(i*2*theta) = ((ixx - iyy) + 1j*(2*ixy))/(ixx + iyy).
```
plazas, line 138: Could you add a few words indicating that the `ixx`, `ixy`, and `iyy` are the second brightness moments? Also, should you use `Ixx` and so on (with a capitalized `I`) here and in the rest of the file as it is used in `ellipticityType` below, for consistency?  (and it might also help to differentiate from the imaginary unit).

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -132,43 +132,79 @@ def __call__(self, df):
...
+    in Bartelmann and Schneider (2001). The other definition differs in the
+    denominator. It is referred to as shear, and denoted by g = (g1, g2) in
+    GalSim and referred to as epsilon-type ellipticity again following the
+    notation in Bartelmann and Schneider (2001). It is defined as
+    E = ((ixx - iyy) + 1j*(2*ixy))/(ixx + iyy + 2sqrt(ixx*iyy - ixy**2)).
 
-    colXx = Field(doc="The column name to get the xx shape component from.",
-                  dtype=str,
-                  default="ixx")
+    Notes
```
plazas, line 147: Do you think it would be useful to add in the notes a short explanation of why we have two definitions, when we want to use one or the other, and why the chi-type is the default in the parameters below? Same comment for the two sizes introduced below.

plazas (approved):
Looks good. I just have minor comments regarding clarifications in the docstrings.

python/lsst/analysis/drp/calcFunctors.py
arunkannawadi: Good catch, Andrés!

python/lsst/analysis/drp/calcFunctors.py
arunkannawadi: In single-visit source tables, the convention has been to call this `ixx`. But for documentation purposes, calling it `Ixx` etc. makes sense.

python/lsst/analysis/drp/calcFunctors.py
```diff
@@ -132,43 +132,79 @@ def __call__(self, df):
...
-
-    For plotting purposes we might want to plot |E|*exp(i*theta).
-    If `halvePhaseAngle` config parameter is set to `True`, then
-    the returned quantity therefore corresponds to |E|*exp(i*theta)
-    """
+    E = |E|exp(i*2*theta) = ((ixx - iyy) + 1j*(2*ixy))/(ixx + iyy).
+    This is sometimes referred to as distortion, and denoted by e = (e1, e2)
+    in GalSim and referred to as chi-type ellipticity following the notation
+    in Bartelmann and Schneider (2001). The other definition differs in the
+    denominator. It is referred to as shear, and denoted by g = (g1, g2) in
```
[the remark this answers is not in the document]
arunkannawadi: I think both `g` and `e` are reduced in the sense, they correspond to elements in a matrix with unit determinant, without any information about the magnification. So I think it makes sense to keep it just as shear here.

python/lsst/analysis/drp/calcFunctors.py
arunkannawadi: I added comments on the two types of definitions, but I'm going to refrain from commenting on the defaults. The defaults are left to whatever they were before this ticket is merged, so that the behavior of the functors called elsewhere remains unchanged. There's no other scientific rationale for the choice of these defaults.

PR #40, opened 2022-05-07
DM-34692: Fix band list for CoaddPlotFlag & Sn Selectors in coaddQAEllip pipeline

laurenam:


python/lsst/analysis/drp/dataSelectors.py
```diff
@@ -1,5 +1,6 @@
-__all__ = ("FlagSelector", "PsfFlagSelector", "StarIdentifier", "GalaxyIdentifier", "UnknownIdentifier",
-           "CoaddPlotFlagSelector", "BaseSNRSelector", "SnSelector")
+__all__ = ("FlagSelector", "PsfFlagSelector", "BaseSNRSelector", "SnSelector",
```
arunkannawadi, line 1: I'm not quite sure about exposing BaseSNRSelector here. It doesn't seem to be used anywhere(?) and if it is used, it would be to define other SNR selectors.

arunkannawadi (approved):
LGTM!

python/lsst/analysis/drp/dataSelectors.py
laurenam: I noticed that it is unused too and plan to address that on another ticket (I first want to check with the original author that removing it is the right thing to do).  Since it was already in the list, I'll leave it as is here and deal with it on the removal ticket.

PR #48, opened 2022-07-05
DM-35336: Several updates and fixes to stellar locus plots

laurenam:


python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -167,30 +234,26 @@ def colorColorFitPlot(self, catPlot, plotInfo, fitParams):
...
-                    The x minimum of the box used in the fit.
-                ``"x2"``
-                    The x maximum of the box used in the fit.
-                ``"y1"``
-                    The y minimum of the box used in the fit.
-                ``"y2"``
-                    The y maximum of the box used in the fit.
+                ``"fitPoints"``
+                    A boolean array indicating which points were usee in the
+                    final ODR fit (`numpy.ndarray` [`bool`]).
```
sr525, line 256: usee -> used

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -229,152 +292,188 @@ def colorColorFitPlot(self, catPlot, plotInfo, fitParams):
...
         # Add some useful information to the plot
-        bbox = dict(alpha=0.9, facecolor="white", edgecolor="none")
-        medMag = np.median(mags)
+        SNsUsedDict = {}
+        if hasattr(self.config.selectorActions, "catSnSelector"):
+            # Compute the effective S/N & mag cuts for the points used in the
+            # fit.
+            for SNBand in self.config.selectorActions.catSnSelector.bands:
+                SNsUsed = (catPlot[SNBand + "_" + plotInfo["SNFlux"]].values[fitPoints]
+                           / catPlot[SNBand + "_" + plotInfo["SNFlux"] + "Err"].values[fitPoints])
```
sr525, line 305: When this gets ported to analysis_tools, there might be functions in place to do this.

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -229,152 +292,188 @@ def colorColorFitPlot(self, catPlot, plotInfo, fitParams):
...
+                fontsize=7, bbox=bbox, va="top")
+        if len(SNsUsedDict) > 0:
+            yLoc = 0.9
+            for SNBand in self.config.selectorActions.catSnSelector.bands:
+                yLoc -= 0.04
+                infoText = "{}: S/N".format(SNBand)
+                infoText += r"$\geq$" + "{:0.1f} [".format(SNsUsedDict[SNBand]["minSnUsed"])
+                infoText += r"$\lesssim$ " + "{:0.1f} mag]".format(SNsUsedDict[SNBand]["medMagUsed"])
+                ax.text(0.04, yLoc, infoText, color="C0", transform=ax.transAxes,
+                        fontsize=6, va="center")
```
sr525, line 331: I don't like the text and will try to come up with something neater for the future.

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -229,152 +292,188 @@ def colorColorFitPlot(self, catPlot, plotInfo, fitParams):
...
         dists = perpDistance(p1, p2, zip(xs[fitPoints], ys[fitPoints]))
 
         # Now we have the information for the perpendicular line we
         # can use it to calculate the points at the ends of the
         # perpendicular lines that intersect at the box edges
-        if np.fabs(fitParams["mHW"]) > 1:
-            xMid = (fitParams["yMin"] - fitParams["bODR2"])/fitParams["mODR2"]
+        if np.fabs(mHW) > 1:
+            xMid = (yMin - fitParams["bODR"])/fitParams["mODR"]
             xs = np.array([xMid - 0.5, xMid, xMid + 0.5])
```
sr525, line 424: Should this be ODR rather than ODR2?

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -229,152 +292,188 @@ def colorColorFitPlot(self, catPlot, plotInfo, fitParams):
...
+        fig.legend(handles=linesForLegend, fontsize=7, loc="lower right", bbox_to_anchor=(0.96, 0.84),
+                   bbox_transform=fig.transFigure, ncol=1)
+
+        axHist.hist(distsScaled, bins=200, histtype="step", label="ODR Fit", color="C0")
+        axHist.hist(distsScaledHW, bins=200, histtype="step", label="HW", color="C0", alpha=0.5)
 
         alphas = [1.0, 0.5]
         handles = [Rectangle((0, 0), 1, 1, color="none", ec="C0", alpha=a) for a in alphas]
-        labels = ["Refit", "HW"]
+        labels = ["ODR Fit", "HW"]
```
sr525, line 476: This should distinguish between the first round of fitting and the second.

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -385,15 +484,18 @@ def colorColorFitPlot(self, catPlot, plotInfo, fitParams):
         minXs = -1*np.min(np.fabs(percsDists))
         maxXs = np.min(np.fabs(percsDists))
         plotPoints = ((dists < maxXs) & (dists > minXs))
-        xs = np.array(dists)[plotPoints]
+        xs = np.array(dists)[plotPoints]*statsUnitScale
```
sr525, line 487: Probably the units should be accounted for in the distance calculation function.

python/lsst/analysis/drp/colorColorFitPlot.py
sr525: Now I see the renaming in the other file, thanks git for making them in that order.

python/lsst/analysis/drp/colorColorFitPlot.py
laurenam: How so?  I re-named `Refit` to `ODR Fit` because the former was never really distinguishing the difference between the first and second rounds of the fit (I was thinking we are only really distinguishing between the "Hard Wired" fit and our final fit on the plots.  Suggestions other than `ODR Fit` welcome (e.g. `Final Fit` or just `Fit`)?

python/lsst/analysis/drp/colorColorFitPlot.py
laurenam: Oh, sure.  I won't change the `perpDistance` function, but I'll do the scaling above where `dists` is computed.

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -75,6 +77,72 @@ class ColorColorFitPlotConfig(pipeBase.PipelineTaskConfig,
...
+        self.yLims = (-0.7, 2.6)
+        self.setConfigDependencies()
+
+    def setConfigDependencies(self):
+        # The following config settings are conditional on other configs.
+        # Set them based on the inter-dependencies here to ensure they are
+        # in sync.  This can (and should) be called in the pipeline definition
+        # if any of the inter-dependent configs are chaned (e.g. self.bands
+        # or self.fluxTypeForColor here.  See plot_wFit_CModel in
+        # stellarLocusPlots.yaml for an example use case.
```
arunkannawadi, line 126: I like this way of setting inter-dependent configs, since I've wondered something similar for rho statistics. I have enforced that they are satisfied in a `validate` method, and I think there should be a `validate` method here as well, to make sure that the config doesn't become invalid by doing something incorrectly after calling. `setConfigDependencies`

laurenam: Heh...it felt totally ugly when I was writing it, but I couldn't think of a better way (I seem to recall slinking into Jim's office asking just how bad this idea was...yes, ugly, but not forbidden!)

TallJimbo (approved):

python/lsst/analysis/drp/colorColorFitPlot.py
laurenam: Just to make sure I understand, are you suggesting this `validate` should happen in `run` (i.e. this is different than the `Config` class `validate`?  I'm pretty sure I tried to make that work, but could never be sure it executed after all possible config manipulations...)?

TallJimbo: No, I'm just saying you should have a config `validate` that checks the invariants instead of relying only on this method having been called to set up the config.  It's the job of whatever is running the task to call the config `validate` and then freeze the config.

python/lsst/analysis/drp/colorColorFitPlot.py
laurenam: Actually, there is nothing being fit "hard wired" wired case.  It is meant to be a sort of "average/reference" fit so that one can identify any field-to-field variations in the location of the stellar locus (particularly in the "P1" principal component).  We compute the mean and standard deviation from the hardwired fit coefficients; a shift in the former would indicate something is "different" in this field (a tract in the current usage), whether it be astrophysical due to a very different stellar population from the one used for the hardwired fit or something amiss with the data reduction.  In the "fit" versions, the mean is ~zero by definition, so this information is lost (it is only the `sigmaMad` of the fit distribution that is the "metric" we typically look at).

I'm open to making a `"hardwired" -> "fixed"` name change, but wouldn't mind punting that to the ticket where this all gets implemented in `analysis_tools` (which did inherit the "hardwired" terminology).  Is that reasonable?

python/lsst/analysis/drp/colorColorFitPlot.py
laurenam: Ah, ok.  I think I am again going to request that this be added in the ticket to update `analysis_tools`, since I would prefer not to have to do all the necessary "validation" (pun intended) of the `validate` function.  Please push back if you don't think that's a reasonable request and want it added on this ticket!

TallJimbo: Definitely; the configuration will certainly have to change somewhat for that transition anyway.

python/lsst/analysis/drp/colorColorFitPlot.py
TallJimbo: Yup, I'm happy to defer any changes to the analysis_tools port.

PR #62, opened 2023-05-09
DM-39031: Use resolved references for GatherResourceUsageTask graph

andy-slac:
This commit adds two command line options to the graph builder script which define output collection names, similarly to `pipetask qgraph` command. Output run is used to resolve references when creating quantum graph for HiPS task. At least one of the `--output` or `--output-runs` options must be provided when invoking the script.

### DIFF
--- a/python/lsst/analysis/drp/gatherResourceUsage.py
+++ b/python/lsst/analysis/drp/gatherResourceUsage.py
@@ -28,17 +28,18 @@ __all__ = (
 import argparse
 from collections import defaultdict
 import dataclasses
+import datetime
 import re
 import logging
-import warnings
 
 import numpy as np
 import pandas as pd
 
-from lsst.daf.butler import Butler, DataCoordinate, DatasetRef, DatasetType, Quantum, UnresolvedRefWarning
+from lsst.daf.butler import Butler, DataCoordinate, DatasetRef, DatasetType, Quantum
 from lsst.daf.butler.core.utils import globToRegex
 from lsst.pex.config import Field, ListField
 from lsst.pipe.base import (
+    Instrument,
     PipelineTask,
     PipelineTaskConfig,
     PipelineTaskConnections,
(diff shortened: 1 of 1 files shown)


python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -331,6 +340,24 @@ def build_quantum_graph_cli(cls, argv):
             default=None,
             help="Data ID expression used when querying for input metadata datasets.",
         )
+        parser.add_argument(
+            "--output",
+            type=str,
+            help="Name of the output CHAINED collection.",
```
timj, line 346: I think this doc string should say that it will create the timestamp run (otherwise people have to read the output-run docs to work it out.)

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -213,6 +215,10 @@ def build_quantum_graph(cls, metadata_refs):
             References to metadata datasets.  Non-metadata datasets are
             silently ignored, in order to make it easier to pass in more
             general dataset query results.
+        output_run : `str`
+            Name of the output RUN collection.
+        graph_metadata : `dict` [ `str`, `Any` ]
+            Graph metadata.
```
timj, line 221: I think either we say that output_run will be added to graph_metadata here (so we can ensure it exists) or else we declare that the output_run must be in metadata and then it doesn't need to be a parameter here. I know this is effectively a private API but it seems wrong to me that we can end up with the API being called without output_run ending up in the metadata.

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -265,23 +271,26 @@ def build_quantum_graph(cls, metadata_refs):
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", category=UnresolvedRefWarning)
```
timj, line 272: All `UnresolvedRefWarning` filters must be removed.

timj (approved):

PR #66, opened 2023-06-07
DM-38952: Switch ConfigurableAction imports to pex_config.

TallJimbo:
@timj , I forgot to include this PR in the rest of the DM-38952 request because I'd forgotten about the branch: it's just updating the import path for some types to correspond to a deprecate-and-move that happened earlier.  It's here just because I got frustrated with all of the warnings while running building QuantumGraphs.

Could you take a quick look?


timj (approved):
Thanks.

PR #70, opened 2023-09-05
DM-38682: Add task for summarizing resource usage into a single table.

TallJimbo:


python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -61,6 +61,107 @@
...
+                df["integrated_runtime"] = ru_table["run_time"].sum()
+
+                quantiles.append(
+                    df[
+                        [
+                            "index",
+                            "quanta",
+                            "task",
+                            "memory",
+                            "init_time",
```
eigerx, line 133: I see that we store `init_time` and `run_time`. Would it be valuable to have a human-readable stop-time also? I'm not sure how useful these quantities are for identifying transient infrastructure errors.

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -61,6 +61,107 @@
...
+            if not input_name.endswith("resource_usage"):
+                continue
+            else:
+                df = ru_table.quantile(
+                    [0.0, 0.01, 0.05, 0.32, 0.50, 0.68, 0.95, 0.99, 1.0],
+                    numeric_only=True,
+                ).reset_index()
+                df["task"] = input_name.replace("_resource_usage", "")
+                df["quanta"] = len(ru_table)
+                df["integrated_runtime"] = ru_table["run_time"].sum()
```
eigerx, line 124: Is the integrated runtime over a workflow, step, graph... ?

python/lsst/analysis/drp/gatherResourceUsage.py
```diff
@@ -61,6 +61,107 @@
...
+                            "task",
+                            "memory",
+                            "init_time",
+                            "run_time",
+                            "integrated_runtime",
+                        ]
+                    ]
+                )
+
+        qq = pd.concat(quantiles)
```
eigerx, line 140: Is `qq` for quantiles? If so a different name might be easier to follow.

eigerx (approved):
Looks good to me!

python/lsst/analysis/drp/gatherResourceUsage.py
TallJimbo: These are actually both durations, not timestamps: `init_time` is the time it takes to do all of the stuff `pipetask` does before it actually call's the task's `runQuantum` method, which amounts to a lot of file-existence checks most of the time. That's the time when things like "Nothing to do for task" or "Skipping already-successful quantum" happen.

python/lsst/analysis/drp/gatherResourceUsage.py
TallJimbo: Integrated over all of the task metadata files it finds in the input collection you give it, so it varies.

python/lsst/analysis/drp/gatherResourceUsage.py
TallJimbo: Renamed to `full_quantiles`.

PR #72, opened 2023-10-09
DM-41043: Drop references to lsst.daf.butler.core.

TallJimbo:
This was not a public namespace, and now it doesn't exist.


timj:
I'm not really sure how my build worked without this patch. Maybe the file isn't imported during the build?

PR #73, opened 2023-10-22
DM-41027: Fix potential infinite loop in analysis_drp colorColorFitPlot

laurenam:


python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -106,6 +107,12 @@ class ColorColorFitPlotConfig(pipeBase.PipelineTaskConfig,
         optional=True,
     )
 
+    minPointsForFit = pexConfig.Field(
+        doc="Minimum number of valid objects to bother attempting a fit.",
+        dtype=int,
+        default=5,
```
taranu, line 113: Does this need to be validated as > 0?

python/lsst/analysis/drp/colorColorFitPlot.py
```diff
@@ -305,10 +330,11 @@ def colorColorFitPlot(self, catPlot, plotInfo, fitParams):
                 SNsUsed = (catPlot[SNBand + "_" + plotInfo["SNFlux"]].values[fitPoints]
```
taranu, line 330: Is it possible for SNsUsed to be empty? I imagine the fit likely would have failed in that case, but if there's a chance it would keep going and result in an infinite loop again, maybe an explicit check would be worthwhile.

taranu (commented):
See comments.

python/lsst/analysis/drp/colorColorFitPlot.py
laurenam: Good idea.  I made it a `RangeField` and gave it `min=1`.

python/lsst/analysis/drp/colorColorFitPlot.py
laurenam: Another good idea.  I put in a check on `len(fitParams["fitPoints"])` above with:
```
if len(fitParams["fitPoints"]) < 1:
    raise ValueError("No fitPoints for {}".format(dataId))
```

